PR #2, opened 2020-02-12
Add KEP process

jbarrick-mesosphere:


kensipe:
review on this tomorrow (thurs)

gerred:
Let's keep it as KUDO Enhancement Proposal for now, and then migrate as necessary, since it's part of the overarching KUDO org/project.

PR #3, opened 2020-02-12
Add unit under test declaration KEP.

jbarrick-mesosphere:


jbarrick-mesosphere:
> Great addition! One potential implentation detail not mentioned here is logging of multiple background commands. The solution should ensure that users can distinguish between the output of different background commands.

Oh yes, that's a great point. I'll add some details here.

jbarrick-mesosphere:
I need to think about that a little bit @nfnt. We're going to want to label the output, something like:

```
[command that is running] log line
```

If we want to the label the output, what do we think would be the best thing to label it with?

* Some index just so they are unique (e.g., `command-1`).
* A truncated form of the command (`kubectl create -f blah..`).
* A user provided label (maybe default to index if not provided?).

The other option might be to colorize each command's output differently.

gerred:
:+1: to merging as draft. @jbarrick-mesosphere if you're OK driving this forward to implementation, work can start on the comments and implementation in the next PR.

PR #4, opened 2020-02-12
Add test composability KEP

jbarrick-mesosphere:


keps/draft-test-composability.md
```diff
@@ -0,0 +1,98 @@
...
+apply:
+- ./install.yaml
+- ../../deploy/
+assert:
+- ../common/check-install.yaml
+error:
+- ./install-error.yaml
+```
+
+This will compose well with the existing `commands` feature of `TestSteps`.
```
nfnt: Can you add a few more details here. We should clarify the order in which `commands` and `apply` are applied. E.g., in `commands` we could run `kudo init` that installs CRDs and `apply` would create resources using these CRDs.

keps/draft-test-composability.md
```diff
@@ -0,0 +1,98 @@
...
+* Provide better integration with existing repository structures (e.g., deploy operator YAMLs from a release directory).
+* Make test file structure easier to understand.
+
+## Proposal
+
+The `TestStep` resource will be extended to allow users to provide manifest, errors, and assert files as a list. Three settings will be added: `apply` , `assert`, `error`. Each is a list of files that can either be a single file name or a directory to recurse over.
+
+* `apply`: a list of files to apply for the test step.
+* `assert`: a list of files containing test asserts to expect.
+* `error`: a list of files containing test errors that will cause the test to fail if their state is observed.
```
ANeumann82: Does it make sense to also have a `delete` resources, to complement the `apply` step?

ANeumann82: Oh, I see there is already a `delete` section...

jbarrick-mesosphere: Yeah, but that would be a good thing to mention here.

kensipe:
rebased

keps/draft-test-composability.md
kensipe: Adding this as a requirement for implementation
Thanks @nfnt you are spot on IMO!

keps/0004-test-composability.md
```diff
@@ -0,0 +1,100 @@
...
+apply:
+- ./install.yaml
+- ../../deploy/
+assert:
+- ../common/check-install.yaml
+error:
+- ./install-error.yaml
+```
+
+This will compose well with the existing `commands` feature of `TestSteps` (note: `commands` will be run prior to either `apply`, `assert`, or `error` during the actual test execution).
```
nfnt: Please also mention the `delete` feature of a `TestStep` and add if this will be run prior to `commands`. Especially for a `TestCase` mentioned below, users may want to clean up deployments at the end of a test case and will use `delete` for that.

keps/0004-test-composability.md
```diff
@@ -0,0 +1,100 @@
...
+* Provide better integration with existing repository structures (e.g., deploy operator YAMLs from a release directory).
+* Make test file structure easier to understand.
+
+## Proposal
+
+The `TestStep` resource will be extended to allow users to provide manifest, errors, and assert files as a list. Three settings will be added: `apply` , `assert`, `error`. Each is a list of files that can either be a single file name or a directory to recurse over.
+
+* `apply`: a list of files to apply for the test step.
+* `assert`: a list of files containing test asserts to expect.
+* `error`: a list of files containing test errors that will cause the test to fail if their state is observed.
```
gpaul: This is cool

justinbarrick:
Hmm, I just had an idea: instead of defining each of these types as different settings (`asserts`, `apply`, `commands`, etc), what if we had one setting that was a list and then users can define their order instead of having to learn it?

Something like:

```
steps:
- assert:
     path: ../assert.yaml 
- command:
     command: kubectl apply
```

That would be both more flexible and more obvious.

keps/0004-test-composability.md
kensipe: @nfnt the `delete` feature should be the same as the others and the `commands` would run prior to it... right?   I can't understand if you are wanting to capture the missing `delete`... or if you are expecting different behavior... or both.   Could you explain any justification other than making sure it is included?

nfnt: I'm not sure what the behavior is right now, that's why we should document the expected ordering here. In my opinion, we should run `delete` first, followed by `commands`, then `apply`, `assert`, `error`. Running `delete` first would ensure that the following features would fail when trying to use resources that just have been deleted. Having a different ordering could create subtle bugs.

kensipe: There is great value in documenting and clearing up kuttl test lifecycle.  I don't think it should hold up this kep.  after a lot of thought... It seems like a separate effort than what this kep is trying to accomplish.

PR #5, opened 2020-02-13
Create KUTTL-as-a-Library KEP

jbarrick-mesosphere:
This is based on discussion in https://github.com/kudobuilder/kudo/issues/1311 as well as the draft KEP from https://docs.google.com/document/d/1PhjMiwG318nPkehDZ4oEMSb6zcKcZeO-/edit#

### DIFF
--- a/keps/0005-kuttl-as-a-library.md
+++ b/keps/0005-kuttl-as-a-library.md
@@ -1,5 +1,5 @@
 ---
-kep-number: 0
+kep-number: 5
 short-desc: Overview of providing KUTTL for use as a Go library
 title: KUTTL as a Library
 authors:
@@ -8,12 +8,10 @@ authors:
 owners:
   - "@mpereira"
   - "@jbarrick-mesosphere"
-editor: TBD
+editor: "@kensipe"
 creation-date: 2020-02-13
-last-updated: 2020-02-13
+last-updated: 2020-02-25
 status: provisional
-see-also:
-  - KEP-2
 ---
 
 # KUTTL as a Library
@@ -55,7 +53,7 @@ Rather than building an entirely new library, we will unify the library and test
 * Err on the side of flexibility
 * Err on the side of returning errors instead of panicking/Fatalf/etc.
 * Cohesive logging output (parameters, formatting, etc.)
-* Separation of concerns between domains (KUDO, Kubernetes, etc.)
+* Separation of concerns between domains (KUDO, Operators, Kubernetes, etc.)
 
 ## Proposal
 
@@ -67,7 +65,7 @@ We will improve and break-down the interface of the kuttl library to make it eas
 * Test setup and tear down methods (is currently coupled to steps).
 * Cluster and test environment provisioning (this is currently coupled to `Harness` in `harness.go`).
 
-Additionally, KUDO-specific helper methods will be provided in the library that can be used by library users.
+Additionally, Operator-specific helper methods will be provided in the library that can be used by library users.
 
 We will also endeavor to stay close to the API used by the kudo-cassandra-operator testing library to ensure that migration of existing tests is easy.
 
--- a/keps/0005-kuttl-as-a-library.md
+++ b/keps/0005-kuttl-as-a-library.md
@@ -5,12 +5,12 @@ title: KUTTL as a Library
 authors:
   - "@mpereira"
   - "@jbarrick-mesosphere"
+  - "@kensipe"
 owners:
-  - "@mpereira"
   - "@jbarrick-mesosphere"
 editor: "@kensipe"
 creation-date: 2020-02-13
-last-updated: 2020-02-25
+last-updated: 2020-04-07
 status: provisional
 ---
 
@@ -53,7 +53,7 @@ Rather than building an entirely new library, we will unify the library and test
 * Err on the side of flexibility
 * Err on the side of returning errors instead of panicking/Fatalf/etc.
 * Cohesive logging output (parameters, formatting, etc.)
-* Separation of concerns between domains (KUDO, Operators, Kubernetes, etc.)
+* Separation of concerns between domains (Operators, Kubernetes, etc.)
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


mpereira:
cc @nfnt

nfnt:
https://github.com/kudobuilder/test-tools was created out of the necessity to have such a library right now. It's already used for testing various D2iQ-maintained operators. It would be great get this functionality in `kuttl`!

keps/draft-kuttl-as-a-library.md
```diff
@@ -0,0 +1,193 @@
...
+* Separation of concerns between domains (KUDO, Kubernetes, etc.)
+
+## Proposal
+
+We will improve and break-down the interface of the kuttl library to make it easier to consume as a downstream developer. Libraries that need to be broken out of kuttl:
+
+* A generic Kubernetes client (currently this is contained inside of `pkg/test/harness.go`).
+* Kubernetes object constructors (this is already pretty good in `pkg/test/utils/kubernetes.go`).
+* Assertion methods (needs better documentation for library usage).
+* Test setup and tear down methods (is currently coupled to steps).
```
nfnt: Wouldn't this, as well as assertions, be something other libraries can provide? Test frameworks like [ginkgo](https://github.com/onsi/ginkgo) provide these functions among others. There's also a lot of assertion libraries, e.g. [testify](https://github.com/stretchr/testify).
Though it would make sense to provide assertion helpers for more complex test scenarios, e.g., asserting that a plan completed `assertPlanCompleted(instance.plan["deploy"])`.

keps/draft-kuttl-as-a-library.md
```diff
@@ -0,0 +1,193 @@
...
+package k8s
+
+func Init(kubectlOptions *kubectl.KubectlOptions) error
+
+func CreateNamespace(namespaceName string) error
+
+func DeleteNamespace(namespaceName string) error
+```
+
+### kudo/kudo.go
```
ANeumann82: The question is, how closely related to KUDO do we want to keep KUTTL? I think we have the plan to remove the KUDO-specific parts from the test harness, and allow kudo to plug into it instead, do we want the same for the library parts?

nfnt: There currently is a need for KUDO-specific part because KUDO doesn't provide a consistent API to use it as a library (somewhat captured in https://github.com/kudobuilder/kudo/issues/1002). While it has a versioned clientset, other functionality, like installing packages is only provided in packages without any compatibility guarantees and should currently be considered internal.

jbarrick-mesosphere: Yeah, while we're decoupling the CLI itself (in order to support more projects), I have no problem with providing helper methods in a special package for kuttl - helpers are cheap. We could also put them in the kudo repo (or even maybe a kudo-client repo?). Not too particular where that part of this goes.

keps/draft-kuttl-as-a-library.md
jbarrick-mesosphere: I'll expand on what I meant in the doc in a bit - but:

* test setup and tear down methods: this would be more along the lines of the namespace creation and deletion methods below - kubernetes-specific test setup. Not about enforcing a test framework / structure or even necessarily a way to run them.
* assertions: what I meant here is more complex assertions: generally kubernetes assertions like you mentioned, but also supporting the same type of asserts we do in the test harness CLI (https://kudo.dev/docs/testing/asserts-errors.html). So, for example, you could use Go to generate a bunch of objects, apply them, and then you can use the assertion method to wait for your expected state instead of having to roll your own thing.

keps/draft-kuttl-as-a-library.md
jbarrick-mesosphere: BTW, KEP for decoupling is here: https://github.com/kudobuilder/kuttl/pull/3

gerred: Agreed. I think the ideal is for `kuttl` to be completely standalone and in-and-of-itself not have any knowledge of KUDO. That's when we succeed with this KEP. For today, if there's cross references, that's OK, but part of this KEP IMO should include a plan to create APIs that specifically detangle the dependency, or at least trigger KUDO to have a kuttl package of it's own for additional helpers.

kensipe:
rebased

keps/0005-kuttl-as-a-library.md
```diff
@@ -0,0 +1,191 @@
...
+package k8s
+
+func Init(kubectlOptions *kubectl.KubectlOptions) error
+
+func CreateNamespace(namespaceName string) error
+
+func DeleteNamespace(namespaceName string) error
+```
+
+### kudo/kudo.go
```
kensipe: I would like to remove this whole section on kudo... this lib should no nothing of kudo IMO

gerred: agreed.

keps/0005-kuttl-as-a-library.md
```diff
@@ -0,0 +1,191 @@
...
+
+* Useful logging by default
+* Depend on existing libraries whenever possible
+* Retrying by default where it makes sense
+* Ease of adding retrying to non-retried functions
+* It is better to be consistent than to be "perfect"
+* Err on the side of flexibility
+* Err on the side of returning errors instead of panicking/Fatalf/etc.
+* Cohesive logging output (parameters, formatting, etc.)
+* Separation of concerns between domains (KUDO, Kubernetes, etc.)
```
kensipe, line 56:
```suggestion
* Separation of concerns between domains (KUDO, Operators, Kubernetes, etc.)
```

keps/0005-kuttl-as-a-library.md
```diff
@@ -0,0 +1,191 @@
...
+
+We will improve and break-down the interface of the kuttl library to make it easier to consume as a downstream developer. Libraries that need to be broken out of kuttl:
+
+* A generic Kubernetes client (currently this is contained inside of `pkg/test/harness.go`).
+* Kubernetes object constructors (this is already pretty good in `pkg/test/utils/kubernetes.go`).
+* Assertion methods (needs better documentation for library usage).
+* Test setup and tear down methods (is currently coupled to steps).
+* Cluster and test environment provisioning (this is currently coupled to `Harness` in `harness.go`).
+
+Additionally, KUDO-specific helper methods will be provided in the library that can be used by library users.
```
kensipe, line 68:
```suggestion
Additionally, Operator-specific helper methods will be provided in the library that can be used by library users.
```

PR #7, opened 2020-02-14
Add KEP for Rego support.

jbarrick-mesosphere:


kensipe:
rebase

jbarrick-mesosphere:
Before moving forward with this, I want to take a look at https://github.com/instrumenta/policies and see if we can be compatible with it.

PR #17, opened 2020-02-25
CLI for kuttl

kensipe:
I'm not sure how this was lost... there was a commit in history that indicates adding the CLI but it is missing.  This is that code.

supports: `help`, `test` and `version`


kensipe:
thanks @ANeumann82

PR #23, opened 2020-03-19
Support Background Processes

kensipe:
In order to support different ways to add a controller under test, we are adding support for running background processes.   This will add the ability to configure something like:

```
commands:
  - command: ./bin/manager 
```
which will run a controller in the background for testing

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sync"
 	"testing"
@@ -43,6 +44,8 @@ type Harness struct {
 	kubeConfigPath string
 	clientLock     sync.Mutex
 	configLock     sync.Mutex
+	stopping       bool
+	bgProcesses    []*exec.Cmd
 }
 
 // LoadTests loads all of the tests in a given directory.
@@ -60,7 +63,7 @@ func (h *Harness) LoadTests(dir string) ([]*Case, error) {
 	tests := []*Case{}
 
 	timeout := h.GetTimeout()
-	h.T.Logf("Going to run test suite with timeout of %d seconds for each step", timeout)
+	h.T.Logf("going to run test suite with timeout of %d seconds for each step", timeout)
 
 	for _, file := range files {
 		if !file.IsDir() {
@@ -207,13 +210,13 @@ func (h *Harness) Config() (*rest.Config, error) {
 	var err error
 
 	if h.TestSuite.StartControlPlane {
-		h.T.Log("Running tests with a mocked control plane (kube-apiserver and etcd).")
+		h.T.Log("running tests with a mocked control plane (kube-apiserver and etcd).")
 		h.config, err = h.RunTestEnv()
 	} else if h.TestSuite.StartKIND {
-		h.T.Log("Running tests with KIND.")
+		h.T.Log("running tests with KIND.")
 		h.config, err = h.RunKIND()
 	} else {
-		h.T.Log("Running tests using configured kubeconfig.")
+		h.T.Log("running tests using configured kubeconfig.")
 		h.config, err = config.GetConfig()
 	}
 
@@ -283,7 +286,9 @@ func (h *Harness) DockerClient() (testutils.DockerClient, error) {
 // RunTests should be called from within a Go test (t) and launches all of the KUDO integration
 // tests at dir.
 func (h *Harness) RunTests() {
-	h.T.Log("Running tests")
+	// cleanup after running tests
+	defer h.Stop()
+	h.T.Log("running tests")
 	tests := []*Case{}
 
 	for _, testDir := range h.TestSuite.TestDirs {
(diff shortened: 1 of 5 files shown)


gerred:
lgtm!

pkg/test/harness.go
```diff
@@ -313,59 +319,64 @@ func (h *Harness) RunTests() {
...
 		}
 	}
-
-	if err := testutils.RunCommands(h.GetLogger(), "default", "", h.TestSuite.Commands, ""); err != nil {
-		h.T.Fatal(err)
+	bgs, errs := testutils.RunCommands(h.GetLogger(), "default", "", h.TestSuite.Commands, "")
+	// assign any background processes first for cleanup if err is fatal
+	h.bgProcesses = append(h.bgProcesses, bgs...)
+	if errs != nil {
+		h.fatal(err)
```
nfnt:
```suggestion
		h.fatal(errs)
```

pkg/test/utils/kubernetes.go
```diff
@@ -952,23 +953,34 @@ func RunCommand(ctx context.Context, namespace string, command string, cmd harne
...
 	if err != nil {
 		if _, ok := err.(*exec.ExitError); ok && cmd.IgnoreFailure {
-			return nil
+			return nil, nil
 		}
+		return nil, err
 	}
 
-	return err
+	if cmd.Background {
```
zmalik: @kensipe do you know if we want to use the same `stdout` and `stderr` as test commands for the background processes?  
Basically using any redirect pipe `>` would be enough to redirect bg process logs to where user wants and not mix them up with test logs?

kensipe: ah... no and I thought I handled that... thanks..  let me check on that

pkg/test/utils/kubernetes.go
```diff
@@ -952,23 +953,34 @@ func RunCommand(ctx context.Context, namespace string, command string, cmd harne
 		fmt.Sprintf("PATH=%s/bin/:%s", actualDir, os.Getenv("PATH")),
 	}
 
-	err = builtCmd.Run()
+	err = builtCmd.Start()
 	if err != nil {
 		if _, ok := err.(*exec.ExitError); ok && cmd.IgnoreFailure {
-			return nil
+			return nil, nil
```
nfnt: Nit: I'd rather have use always return `builtCmd`, even if it's run in foreground. This way the return value of this function isn't dependent on some options of its input. Let's have the caller decide whether they need the reference to the process or not.

pkg/test/utils/kubernetes.go
```diff
@@ -952,23 +953,34 @@ func RunCommand(ctx context.Context, namespace string, command string, cmd harne
...
 	}
 
-	return err
+	if cmd.Background {
+		return builtCmd, nil
+	}
+
+	err = builtCmd.Wait()
+	if _, ok := err.(*exec.ExitError); ok && cmd.IgnoreFailure {
+		return nil, nil
```
nfnt: See above.

pkg/test/utils/kubernetes.go
```diff
@@ -952,23 +953,34 @@ func RunCommand(ctx context.Context, namespace string, command string, cmd harne
...
+	}
+	return nil, err
 }
 
 // RunCommands runs a set of commands, returning any errors.
 // If `command` is set, then `command` will be the command that is invoked (if a command specifies it already, it will not be prepended again).
-func RunCommands(logger Logger, namespace string, command string, commands []harness.Command, workdir string) []error {
+// commands running in the background are returned
```
nfnt: No, that isn't quite right. For commands running in foreground, `bgs` will have a `nil` entry. While this is a subtle difference, it can be a source of errors when mixing foreground and background in commands. Let's either be clear about this and have the caller distinguish between `nil` and non-`nil` entries or return `exec.Cmd` for all commands (as suggested for `RunCommand`) or only return `exec.Cmd` for background tasks.

pkg/test/harness.go
```diff
@@ -313,59 +319,64 @@ func (h *Harness) RunTests() {
...
 		if _, err := testutils.InstallManifests(context.TODO(), cl, dClient, manifestDir); err != nil {
-			h.T.Fatal(err)
+			h.fatal(err)
 		}
 	}
-
-	if err := testutils.RunCommands(h.GetLogger(), "default", "", h.TestSuite.Commands, ""); err != nil {
-		h.T.Fatal(err)
+	bgs, errs := testutils.RunCommands(h.GetLogger(), "default", "", h.TestSuite.Commands, "")
+	// assign any background processes first for cleanup if err is fatal
```
zen-dog:
```suggestion
	// assign any background processes first for cleanup if err is fatal. should there be any errors
	// Stop method (deferred above) will clean up any dangling background processes.
```

pkg/test/harness.go
```diff
@@ -392,6 +403,15 @@ func (h *Harness) Stop() {
 		return
 	}
 
+	if h.bgProcesses != nil {
+		for _, process := range h.bgProcesses {
```
zen-dog: nit:
```suggestion
		for _, p := range h.bgProcesses {
```

pkg/test/harness.go
```diff
@@ -415,6 +435,17 @@ func (h *Harness) Stop() {
 	}
 }
 
+// wraps Test.Fatal in order to clean up harness
+func (h *Harness) fatal(args ...interface{}) {
+	// clean up on fatal in setup
+	if !h.stopping {
+		// stopping prevents reentry into h.Stop
+		h.stopping = true
```
zen-dog: I guess, `h.fatal()` is never meant to be called from a goroutine? Please add a comment to the method description

pkg/test/utils/kubernetes.go
zen-dog: Yeah, intermingled logs are no help.

kensipe: totally agree!  It was a mistake for sure... thanks for the eagle eyes

pkg/test/harness.go
kensipe: damn good eye!

pkg/test/utils/kubernetes.go
kensipe: thank you!  made this statement correct... it is what was expect... a slice of only bg

pkg/test/utils/kubernetes.go
kensipe: `builtCmd` is a `*exec.Cmd` which doesn't have the necessary metadata... we would have to mod `harness.Command` to have this pointer and return `harness.Command` from this func.... worth considering...  I don't get it though...  The command is run and exited... perhaps you can reason about exit codes from the calling code (which was not previously needed and we don't have a requirement for it).

nfnt: Makes sense, thanks for the explanation.

pkg/test/utils/kubernetes_integration_test.go
```diff
@@ -107,3 +110,56 @@ func TestClientWatch(t *testing.T) {
 
 	events.Stop()
 }
+
+func TestRunCommand(t *testing.T) {
```
zen-dog: Would you please add tests making sure that hanging background processes are cleaned up at the end?

PR #24, opened 2020-03-20
remove default ns for kubectl commands

zmalik:
I have a use case where I need to create resources from a yaml file and those resources belong to different namespaces. So I cannot use `--namespace` flag. But defaulting to `default` flag also doesn't work in this case

Example:
```
apiVersion: kudo.dev/v1alpha1
kind: TestSuite
kubectl:
- apply -f https://mesosphere.github.io/kubeaddons/bundle.yaml
testDirs:
- ./repository/zookeeper/tests
startKIND: true
startKUDO: false
timeout: 300
```
this ends up with errors:
```
the namespace from the provided object "kubeaddons" does not match the namespace "default". You must pass '--namespace=kubeaddons' to perform this operation.
the namespace from the provided object "kubeaddons" does not match the namespace "default". You must pass '--namespace=kubeaddons' to perform this operation.
the namespace from the provided object "kube-system" does not match the namespace "default". You must pass '--namespace=kube-system' to perform this operation.
```

I think we should use namespaces for test cases but we shouldn't use the `default` namespace for the `kubectl commands`. Instead, ask the user to define it through a `--namespace` flag or use the one defined in the definition of the objects in yaml files. 

Signed-off-by: Zain Malik <zmalikshxil@gmail.com>

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -371,7 +371,7 @@ func (h *Harness) Setup() {
 		h.fatal(errs)
 	}
 
-	if errs := testutils.RunKubectlCommands(h.GetLogger(), "default", h.TestSuite.Kubectl, ""); errs != nil {
+	if errs := testutils.RunKubectlCommands(h.GetLogger(), "", h.TestSuite.Kubectl, ""); errs != nil {
 		h.fatal(errs)
 	}
 }
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -920,7 +920,7 @@ func GetArgs(ctx context.Context, command string, cmd harness.Command, namespace
 			return nil, err
 		}
 
-		if *namespaceParsed == "" {
+		if *namespaceParsed == "" && namespace != "" {
 			argSlice = append(argSlice, "--namespace", namespace)
 		}
 	}
--- a/pkg/test/utils/kubernetes_test.go
+++ b/pkg/test/utils/kubernetes_test.go
@@ -404,6 +404,14 @@ func TestGetKubectlArgs(t *testing.T) {
 				"kubectl", "kudo", "test", "--namespace", "default",
 			},
 		},
+		{
+			testName:  "namespace not present isn't appended",
+			namespace: "",
+			args:      "kubectl kudo test",
+			expected: []string{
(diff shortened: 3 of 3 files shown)


zmalik:
so we are going to remove kubectl command step?

zmalik:
the command step also applies the same `default` namespace and doesn't work
```
    kuttl: logger.go:37: 10:38:46 |  | Running command: kubectl {apply -f https://mesosphere.github.io/kubeaddons/bundle.yaml %!s(bool=true) %!s(bool=false) %!s(bool=false)}
    kuttl: logger.go:37: 10:38:47 |  | the namespace from the provided object "kubeaddons" does not match the namespace "default". You must pass '--namespace=kubeaddons' to perform this operation.
        the namespace from the provided object "kubeaddons" does not match the namespace "default". You must pass '--namespace=kubeaddons' to perform this operation.
        the namespace from the provided object "kube-system" does not match the namespace "default". You must pass '--namespace=kube-system' to perform this operation.
```
I think we should probably fix this before the first release of `kuttl` instead of providing backward compatibility to released before the first release

justinbarrick:
Yeah, we should remove the kubectl setting. The commands setting only appends a namespace if you set `namespaced: true`. https://github.com/kudobuilder/kuttl/blob/13c66eb2141638dababe29539a948c6c24402d17/pkg/test/utils/kubernetes.go#L914

What was your configuration?

zmalik:
let me check if there is some default value that is setting this. I didn't set the `namespaced: true` for command step
```
apiVersion: kudo.dev/v1alpha1
kind: TestSuite
commands:
  - command: kuebctl apply -f https://mesosphere.github.io/kubeaddons/bundle.yaml
testDirs:
- ./repository/zookeeper/tests
startKIND: true
startKUDO: false
timeout: 300
```

zmalik:
was still using old tests for some reason. using command works for multi-ns resources in a `yaml` bundle. 
that works @justinbarrick thank you!
@kensipe @justinbarrick I'm closing this PR and can help with documentation around the command vs kubectl command and would vote to totally exclude `kubectl` command from any examples we put in this repo.

justinbarrick:
Thank you! I’ll go ahead and drop the setting before we cut our first release too.

PR #35, opened 2020-03-24
Fixes Background Process Leaking Bug

### LINKED ISSUE #33: Kill of Background Processes Failing
The newly added technique to start a background process is working in testing but leaves the process running after exit.   It isn't clear yet why... theories:
1. the process exits too quickly for the kill to take place
2. the process is a grandchild (not a child) which go doesn't kill (frankly we may need to support killing all child and grandchild processes anyway.

for ref: https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773

kensipe:
Fixes #33 

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -409,6 +409,13 @@ func (h *Harness) Stop() {
 			if err != nil {
 				h.T.Log("background process kill error", err)
 			}
+			ps, err := p.Process.Wait()
+			if err != nil {
+				h.T.Log("background process kill wait error", err)
+			}
+			if ps != nil {
+				h.T.Log("background process exit code ", ps.ExitCode())
+			}
 		}
 	}
 
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -404,17 +404,17 @@ func (h *Harness) Stop() {
 
 	if h.bgProcesses != nil {
 		for _, p := range h.bgProcesses {
-			h.T.Log("killing process ", p)
+			h.T.Logf("killing process %q", p)
 			err := p.Process.Kill()
 			if err != nil {
-				h.T.Log("background process kill error", err)
+				h.T.Logf("bg process: %q kill error %v", p, err)
 			}
 			ps, err := p.Process.Wait()
 			if err != nil {
-				h.T.Log("background process kill wait error", err)
+				h.T.Logf("bg process: %q kill wait error %v", p, err)
 			}
 			if ps != nil {
-				h.T.Log("background process exit code ", ps.ExitCode())
+				h.T.Logf("bg process: %q exit code %v", p, ps.ExitCode())
 			}
 		}
 	}
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -317,6 +317,7 @@ func (h *Harness) RunTests() {
 			})
 		}
 	})
+	h.T.Log("run tests finished")
 }
 
 // Run the test harness - start the control plane and then run the tests.
@@ -333,11 +334,13 @@ func (h *Harness) Setup() {
 
 	cl, err := h.Client(false)
 	if err != nil {
+		h.T.Log("fatal error getting client")
 		h.fatal(err)
 	}
 
 	dClient, err := h.DiscoveryClient()
 	if err != nil {
+		h.T.Log("fatal error getting discovery client")
 		h.fatal(err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/test/harness.go
```diff
@@ -393,6 +402,16 @@ func (h *Harness) Stop() {
...
+	if h.bgProcesses != nil {
+		for _, p := range h.bgProcesses {
+			h.T.Log("killing process ", p)
+			err := p.Process.Kill()
+			if err != nil {
+				h.T.Log("background process kill error", err)
+			}
+		}
+	}
+
```
kensipe: This needed to move up... the case below (which previously came before could return) leaking back processes

pkg/test/utils/kubernetes.go
```diff
@@ -1006,24 +1006,24 @@ func RunCommands(logger Logger, namespace string, command string, commands []har
...
-			}
 		}
-
-		logger.Log(stderr.String())
-		logger.Log(stdout.String())
+		if bg != nil {
+			bgs = append(bgs, bg)
+		} else {
+			logger.Log(stderr.String())
+			logger.Log(stdout.String())
```
kensipe: no logging if background

pkg/test/utils/kubernetes.go
```diff
@@ -1006,24 +1006,24 @@ func RunCommands(logger Logger, namespace string, command string, commands []har
...
 		if err != nil {
 			errs = append(errs, err)
-			if bg != nil {
-				bgs = append(bgs, bg)
-			}
 		}
-
-		logger.Log(stderr.String())
-		logger.Log(stdout.String())
+		if bg != nil {
```
kensipe: always add bg (this was the main bug)

pkg/test/harness.go
```diff
@@ -393,6 +402,16 @@ func (h *Harness) Stop() {
 		}
 	}
 
+	if h.bgProcesses != nil {
```
zen-dog: Fun fact, I've asked for a test for this case in the original #23 PR where this [bug was introduced](https://github.com/kudobuilder/kuttl/pull/23#discussion_r395609070) ;)

kensipe: @zen-dog definitely!   added!

pkg/test/harness.go
```diff
@@ -393,6 +402,16 @@ func (h *Harness) Stop() {
...
+		for _, p := range h.bgProcesses {
+			h.T.Log("killing process ", p)
+			err := p.Process.Kill()
+			if err != nil {
+				h.T.Log("background process kill error", err)
+			}
```
zen-dog: Shouldn't we `Wait()` here for the process? 
```Wait waits for the Process to exit, and then returns a ProcessState describing its status and an error, if any. Wait releases any resources associated with the Process. On most operating systems, the Process must be a child of the current process or an error will be returned.```

File: pkg/test/harness_test.go [hunk 6]
```diff
@@ -81,3 +84,29 @@ func TestAddNodeCaches(t *testing.T) {
...
+	defer h.Stop()
+
+	// setup creates bg processes
+	assert.Equal(t, 1, len(h.bgProcesses))
+	// process is alive
+	assert.NoError(t, h.bgProcesses[0].Process.Signal(syscall.Signal(0)))
+
+	// cleans up bg processes
+	h.Stop()
+	_, err := h.bgProcesses[0].Process.Wait()
```
zen-dog: See my comment above about waiting

pkg/test/harness.go
kensipe: @zen-dog I read the docs...  I'm not sure why we would want to wait... it doesn't add any value at this point

pkg/test/harness_test.go
kensipe: this wait is to ensure we don't have flaky tests

pkg/test/harness.go
kensipe: The termination of the test harness does not need to wait for other processes outside of its control to exit before exiting.   The harness is killing as we would expect.   Do we want the harness to hang because a user defined process is hanging?     If we were to `wait()` I would rewrite this such that the kill and waits are concurrent and wait for all routines to return.  this seems like a lot of complexity for no gain IMO

pkg/test/harness.go
```diff
@@ -393,6 +402,23 @@ func (h *Harness) Stop() {
 		}
 	}
 
+	if h.bgProcesses != nil {
+		for _, p := range h.bgProcesses {
+			h.T.Log("killing process ", p)
+			err := p.Process.Kill()
+			if err != nil {
+				h.T.Log("background process kill error", err)
```
zen-dog: Please log the process name in all places (L414 and L417)  (are command and arguments also available for logging?)

pkg/test/harness.go
```diff
@@ -393,6 +402,23 @@ func (h *Harness) Stop() {
 		}
 	}
 
+	if h.bgProcesses != nil {
+		for _, p := range h.bgProcesses {
+			h.T.Log("killing process ", p)
+			err := p.Process.Kill()
+			if err != nil {
+				h.T.Logf("bg process: %v kill error %v", p, err)
```
zen-dog:
```suggestion
				h.T.Logf("bg process: %q kill error %v", p, err)
```

zen-dog: and the others too

PR #49, opened 2020-04-02
Log output of background commands.

jbarrick-mesosphere:
We add a Write() method to the Logger interface to implement the
io.Writer interface and then the logger is passed to the commands as
stdout/stderr.

Signed-off-by: jbarrick@mesosphere.com <jbarrick@mesosphere.com>

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -961,10 +961,8 @@ func RunCommand(ctx context.Context, namespace string, command string, cmd harne
 	}
 
 	builtCmd.Dir = cwd
-	if !cmd.Background {
-		builtCmd.Stdout = stdout
-		builtCmd.Stderr = stderr
-	}
+	builtCmd.Stdout = stdout
+	builtCmd.Stderr = stderr
 	builtCmd.Env = []string{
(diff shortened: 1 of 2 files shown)


jbarrick-mesosphere:
@kensipe the test failed for some reason (looks to be the linter used too much memory), but I don't have permission to re-run

kensipe:
also... a number of re-runs show the same result... how is it possible that `golangci-lint` requires more than 4GB?   I'm finding a lot of examples of this linter using > 5GB.  I'm not sure why kudo isn't failing in this way.   researching but isn't blocking... local lint looks fine.

kensipe:
rebased to master after master had linter fix land

kensipe:
after some thought... I like this... but I want to add an attribute to `Command` or connecting to stdout/err
This is great

kensipe:
it is also likely a benefit to prepend log with process info... name or pid or ... so those logs can be filtered

jbarrick-mesosphere:
Both of those ideas sound great :+1: it's also worth noting that the logs are buffered and organized by process so they won't be intermingled.

I'll go ahead and update this PR with those changes.

PR #65, opened 2020-04-22
Test Clean up

kensipe:
unit tests should not have a need on external services like kind.  
refactored to move from unit to integration


gerred:
lgtm!

PR #72, opened 2020-05-04
ENV VAR shell expansion to KUTTL commands

### LINKED ISSUE #71: Using variables in test steps
Support for using variables inside the KUTTL steps. Right now we can use env variables in command steps. But it would help if we could extend that to all test steps. Using env variables or other ways of providing variables. 

**Why is this needed**:
```
apiVersion: kudo.dev/v1beta1
kind: Instance
metadata:
  name: cassandra
spec:
  name: "cassandra"
  parameters:
    NODE_DOCKER_IMAGE: mesosphere/cassandra:<generated-value>
```
for example when updating a Kubernetes object as a step,  like in the above example, sometimes we need to add a runtime value to the test. In the above case the `<generated-value>`, which could come from env variables or from a file.

kensipe:
Looking for input on this!  This specifically solves an issue with D2iQ team where output of a CI build provides the "name" or "version" of an image which is to be used under test.  This seems like an easy solution / use case.

Concerns are that this will lead to less portable KUTTL test scripts where awareness is necessary outside of the test suite.   The consensus between @zmalik and I is that there are different categories of tests (which was true before but was implied).  For instance... there are tests which work under 1 test control plane and not another.   It is also valuable to use KUTTL for setup and variation for Mixed Workload Testing for instance for which these tests are not necessarily designed to be portable to other platform environments.

The Acceptance Criteria used to build feature:

##  Expansion of commands
`VER=0.2.0 kuttl test some/test/folder`

With a TestStep has
```
apiVersion: kudo.dev/v1beta1
kind: TestStep
commands:
  - command: kubectl apply -f ../foo/script-$VER.sh
```

## Script Access to ENV
The last commit adds ability for command script to access ENV VARs

TestStep
```
apiVersion: kudo.dev/v1beta1
kind: TestStep
commands:
  - command: ./example.sh
```

```
#!/bin/bash

echo image is $IMAGE_TAG
```
Allowing for `IMAGE_TAG=foo/0.2.0 kuttl test some/test/folder`

Fixes #71

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -912,16 +912,19 @@ func StartTestEnvironment() (env TestEnvironment, err error) {
 }
 
 // GetArgs parses a command line string into its arguments and appends a namespace if it is not already set.
+// provides OS expansion of defined ENV VARs inside args to commands.  The expansion is limited to what is defined on the OS
+// and not variables defined for kuttl tests
 func GetArgs(ctx context.Context, command string, cmd harness.Command, namespace string) (*exec.Cmd, error) {
 	argSlice := []string{}
 
-	argSplit, err := shlex.Split(cmd.Command)
+	c := os.ExpandEnv(cmd.Command)
+	argSplit, err := shlex.Split(c)
 	if err != nil {
 		return nil, err
 	}
 
 	if command != "" && argSplit[0] != command {
-		argSlice = append(argSlice, command)
+		argSlice = append(argSlice, os.ExpandEnv(command))
 	}
 
 	argSlice = append(argSlice, argSplit...)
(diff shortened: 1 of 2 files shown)


gerred:
Is there a way to escape this to still use literals? This will be important for pod commands that take env vars from a ConfigMap, Secret, or literal.

It might be worth considering changing the syntax over to something like ${MYENVVAR} so we can leave Deployment specs alone, since that's equally hard to change in Kubernetes itself, and might cause issues with kuttl trying to expand something that should be a string literal, yeah?

gerred:
Discussed in standup this only affects TestStep, so this looks good.

jbarrick-mesosphere:
:partying_face:

PR #77, opened 2020-05-07
 Enable most admission controllers for the control plane tests

zen-dog:
Summary:
we used to be very permissive with admission controllers for the control plane tests by using `--admission-control=AlwaysAdmit` option. It is deprecated and it effectively disables all admission controllers. This PR introduces a much narrower setup by disabling only two admission controllers: `--disable-admission-plugins=ServiceAccount,NamespaceLifecycle` which are still needed for some tests.

Signed-off-by: Aleksey Dukhovniy <alex.dukhovniy@googlemail.com>

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -187,7 +187,7 @@ func (h *Harness) addNodeCaches(dockerClient testutils.DockerClient, kindCfg *ki
 func (h *Harness) RunTestEnv() (*rest.Config, error) {
 	started := time.Now()
 
-	testenv, err := testutils.StartTestEnvironment()
+	testenv, err := testutils.StartTestEnvironment(h.TestSuite.ControlPlaneArgs)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -191,7 +192,9 @@ func (h *Harness) RunTestEnv() (*rest.Config, error) {
 		return nil, err
 	}
 
-	h.T.Log("started test environment (kube-apiserver and etcd) in", time.Since(started))
+	h.T.Logf("started test environment (kube-apiserver and etcd) in %v, with following options:\n%s",
+		time.Since(started),
+		strings.Join(testenv.Environment.KubeAPIServerFlags, "\n"))
 	h.env = testenv.Environment
 
 	return testenv.Config, nil
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -70,7 +70,7 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--admission-control=AlwaysAdmit",
+	"--disable-admission-plugins=ServiceAccount,NamespaceLifecycle",
 	"--service-cluster-ip-range=10.0.0.0/24",
 }
 
@@ -891,13 +891,7 @@ func StartTestEnvironment() (env TestEnvironment, err error) {
 		KubeAPIServerFlags: append(APIServerDefaultArgs, "--advertise-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}"),
 	}
 
-	// Retry up to three times for the test environment to start up in case there is a port collision (#510).
-	for i := 0; i < 3; i++ {
-		env.Config, err = env.Environment.Start()
-		if err == nil {
-			break
-		}
-	}
+	env.Config, err = env.Environment.Start()
 
 	if err != nil {
 		return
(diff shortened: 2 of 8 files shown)


go.mod
```diff
@@ -16,7 +16,6 @@ require (
 	github.com/gorilla/mux v1.6.2 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
-	github.com/kudobuilder/kudo v0.11.1 // indirect
```
zen-dog, line 19: `go mod tidy` 🤷

pkg/test/utils/kubernetes.go
```diff
@@ -69,7 +69,7 @@ var APIServerDefaultArgs = []string{
 	"--insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}",
 	"--insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 	"--secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}",
-	"--admission-control=AlwaysAdmit",
+	"--disable-admission-plugins=ServiceAccount,NamespaceLifecycle",
```
zen-dog, line 72: This is the main change. I also thought about exposing `APIServerDefaultArgs` to the `kuttl` users (different tests might require different admission controller set) but wrapping and wiring these through `kuttl` turned out to be tedious so I just edited it here. Especially, because these are only applied for the controller plane tests.

pkg/test/utils/kubernetes.go
```diff
@@ -890,13 +890,7 @@ func StartTestEnvironment() (env TestEnvironment, err error) {
...
 	}
 
-	// Retry up to three times for the test environment to start up in case there is a port collision (#510).
-	for i := 0; i < 3; i++ {
-		env.Config, err = env.Environment.Start()
-		if err == nil {
-			break
-		}
-	}
+	env.Config, err = env.Environment.Start()
```
zen-dog, line 893: I don't think this loop makes sense anymore. Envtest already has [logic to retry](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/server.go#L270) starting the controller plane.

pkg/test/harness.go
```diff
@@ -191,7 +192,9 @@ func (h *Harness) RunTestEnv() (*rest.Config, error) {
 		return nil, err
 	}
 
-	h.T.Log("started test environment (kube-apiserver and etcd) in", time.Since(started))
+	h.T.Logf("started test environment (kube-apiserver and etcd) in %v, with following options:\n%s",
+		time.Since(started),
+		strings.Join(testenv.Environment.KubeAPIServerFlags, "\n"))
```
kensipe, line 197: output for other reviewers...

```
 kuttl: harness.go:195: started test environment (kube-apiserver and etcd) in 4.597689123s, with following options:
        --advertise-address=127.0.0.1
        --etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}
        --cert-dir={{ .CertDir }}
        --insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}
        --insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
        --secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}
        --disable-admission-plugins=ServiceAccount,NamespaceLifecycle
        --service-cluster-ip-range=10.0.0.0/24
        --advertise-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
```

pkg/test/utils/kubernetes.go
kensipe: it seems that you captured the concerns 
1) only affects mock control plane
2) and I also thought exposing thru testsuite makes sense... but it seems fine to capture that in another PR if we gather consensus around it.

I'm concerned with making sure that many use cases are supported... for which I think expose to end user is necessary.  perhaps we can take the route we take with kind config...  where the full `APIServerDefaultArgs` can be provided in a file and can be reference as an override in testsuite and a flag.   for expert users.

pkg/test/utils/kubernetes.go
```diff
@@ -1004,7 +998,7 @@ func RunCommands(logger Logger, namespace string, command string, commands []har
 	}
 
 	for _, cmd := range commands {
-		logger.Logf("running command: %s %s", command, cmd)
+		logger.Logf("running command: %s %q", command, cmd.Command)
```
kensipe, line 1001: output for other reviewers

```
kuttl/harness/cli-test: logger.go:41: 16:24:57 | cli-test/1-patch | running command:  "kubectl label pod cli-test-pod test=true"
```

vs what used to be:

```
kuttl/harness/cli-test: logger.go:41: 16:33:29 | cli-test/1-patch | running command:  {"kubectl label pod cli-test-pod test=true" %!q(bool=true) %!q(bool=false) %!q(bool=false)}
```

pkg/test/utils/kubernetes.go
zen-dog: 💯 valid points, which you already addressed in #78 Let's continue the discussion there 👍

pkg/test/utils/kubernetes.go
ANeumann82: Oh thanks! That's a lot cleaner :D

kensipe:
The approvals still stand... we had a PR on a PR both with all needed approvals.    The merge of the first PR reset approvals but we will merge without forcing more attention after tests pass.

PR #79, opened 2020-05-07
make controller-runtime configuration available to advance users

kensipe:
This is based on the PR #77 from @zen-dog and extends the ability to configure the api server for advanced users.   There is an example configuration in the new `config` folder.

Example use is:

`go run cmd/kubectl-kuttl/main.go test pkg/test/test_data/ --control-plane-config=config/example-mock-controller.config `

An output for a run without configuration looks like:
```
go run cmd/kubectl-kuttl/main.go test pkg/test/test_data/ 
=== RUN   kuttl
    kuttl: harness.go:336: starting setup
    kuttl: harness.go:216: running tests with a mocked control plane (kube-apiserver and etcd).
    kuttl: harness.go:195: started test environment (kube-apiserver and etcd) in 4.506242639s, with following options:
        --advertise-address=127.0.0.1
        --etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}
        --cert-dir={{ .CertDir }}
        --insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}
        --insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
        --secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}
        --disable-admission-plugins=ServiceAccount,NamespaceLifecycle
        --service-cluster-ip-range=10.0.0.0/24
        --advertise-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
```

An output with configuration looks like:
```
go run cmd/kubectl-kuttl/main.go test pkg/test/test_data/ --control-plane-config=config/example-mock-controller.config 
=== RUN   kuttl
    kuttl: harness.go:336: starting setup
    kuttl: harness.go:216: running tests with a mocked control plane (kube-apiserver and etcd).
    kuttl: harness.go:195: started test environment (kube-apiserver and etcd) in 4.833250298s, with following options:
        --etcd-servers={{ if .EtcdURL }}{{ .EtcdURL.String }}{{ end }}
        --cert-dir={{ .CertDir }}
        --insecure-port={{ if .URL }}{{ .URL.Port }}{{ end }}
        --insecure-bind-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
        --secure-port={{ if .SecurePort }}{{ .SecurePort }}{{ end }}
        --admission-control=AlwaysAdmit
        --service-cluster-ip-range=10.0.0.0/24
        --advertise-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}
```

Fixes #

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -47,6 +47,7 @@ func newTestCmd() *cobra.Command {
 	skipClusterDelete := false
 	parallel := 0
 	artifactsDir := ""
+	mockControllerFile := ""
 
 	options := harness.TestSuite{}
 
@@ -151,6 +152,17 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 			if len(options.TestDirs) == 0 {
 				return errors.New("no test directories provided, please provide either --config or test directories on the command line")
 			}
+			var APIServerArgs []string
+			var err error
+			if mockControllerFile != "" {
+				APIServerArgs, err = testutils.ReadMockControllerConfig(mockControllerFile)
+			} else {
+				APIServerArgs = testutils.APIServerDefaultArgs
+			}
+			if err != nil {
+				return err
+			}
+			options.ControlPlaneArgs = APIServerArgs
 
 			return nil
 		},
@@ -171,6 +183,7 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 	testCmd.Flags().StringSliceVar(&manifestDirs, "manifest-dir", []string{}, "One or more directories containing manifests to apply before running the tests.")
 	testCmd.Flags().StringVar(&testToRun, "test", "", "If set, the specific test case to run.")
 	testCmd.Flags().BoolVar(&startControlPlane, "start-control-plane", false, "Start a local Kubernetes control plane for the tests (requires etcd and kube-apiserver binaries, cannot be used with --start-kind).")
+	testCmd.Flags().StringVar(&mockControllerFile, "control-plane-config", "", "Path to file to load controller-runtime APIServer configuration arguments (only usedful when --startControlPlane).")
 	testCmd.Flags().BoolVar(&startKIND, "start-kind", false, "Start a KIND cluster for the tests (cannot be used with --start-control-plane).")
 	testCmd.Flags().StringVar(&kindConfig, "kind-config", "", "Specify the KIND configuration file path (implies --start-kind, cannot be used with --start-control-plane).")
 	testCmd.Flags().StringVar(&kindContext, "kind-context", "", "Specify the KIND context name to use (default: kind).")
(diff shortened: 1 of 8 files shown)


zen-dog:
Nice! But do we need a new config file for these args? Or put it into `kudo-test.yaml`? I mean, we already have all the other options there e.g. `startControlPlane: true`. Feels like another section like `controlPlaneArgs` would make sense there. Wdyt?

pkg/kuttlctl/cmd/test.go
```diff
@@ -151,6 +152,17 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 			if len(options.TestDirs) == 0 {
 				return errors.New("no test directories provided, please provide either --config or test directories on the command line")
 			}
+			var APIServerArgs []string
+			var err error
+			if mockControllerFile != "" {
+				APIServerArgs, err = testutils.ReadMockControllerConfig(mockControllerFile)
```
ANeumann82, line 158: I'd like to see a log that we're using a user provided config here, and which one, it's easy to overlook things like that when configuring kuttl for tests.

pkg/kuttlctl/cmd/test.go
```diff
@@ -171,6 +183,7 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 	testCmd.Flags().StringSliceVar(&manifestDirs, "manifest-dir", []string{}, "One or more directories containing manifests to apply before running the tests.")
 	testCmd.Flags().StringVar(&testToRun, "test", "", "If set, the specific test case to run.")
 	testCmd.Flags().BoolVar(&startControlPlane, "start-control-plane", false, "Start a local Kubernetes control plane for the tests (requires etcd and kube-apiserver binaries, cannot be used with --start-kind).")
+	testCmd.Flags().StringVar(&mockControllerFile, "control-plane-config", "", "Path to file to load controller-runtime APIServer configuration arguments (only usedful when --startControlPlane).")
```
ANeumann82, line 186:
```suggestion
	testCmd.Flags().StringVar(&mockControllerFile, "control-plane-config", "", "Path to file to load controller-runtime APIServer configuration arguments (only useful when --startControlPlane).")
```

pkg/kuttlctl/cmd/test.go
kensipe: @ANeumann82 there is already an stdout output of the exact configuration used at the point of starting the control plane... is that sufficient or were you looking for more than that?

kensipe:
@ANeumann82 I agree that the templating is 👎 in the config... but this an api-server control and is only for advance users... my hope is no one needs it... if they do... they have full control and there be dragons :)

pkg/kuttlctl/cmd/test.go
ANeumann82: I would like to see the actual file name of the config file that is used, but we don't have the `*testing.T` here which means the logging would be a bit weird. 

I'm ok with the way it is.

PR #81, opened 2020-05-08
Add namespace to ENV variables when a command is executed

ANeumann82:
We might have to discuss this, but this just came up when using a script to modify the cluster in a test. 

I was able to use a workaround (set `namespaced: true` and then parse the added `--namespace <ns-name>` but it feels pretty clumsy, having the Namespace in the environment makes a lot of sense to me.


ANeumann82:
I can't approve my own PR, but looks good to me

PR #95, opened 2020-05-15
Update kind dependency to 0.8.4

ANeumann82:

- Updates KIND dependency to 0.8.1
- Updates K8s dependencies to 0.18.2 (required by newer KIND)

We're currently investigating an issue in KUDO C* that may be related to https://github.com/kubernetes-sigs/kind/issues/1288

It seems that KIND 0.8.1 fixes this issue.

Fixes #


zmalik:
@kensipe mentioned some issues with `0.8.0` I wonder if those are fixed with `0.8.1`
https://kubernetes.slack.com/archives/CEKK1KTN2/p1589316603314000

ANeumann82:
No, they are not yet fixed, as far as I know. That's why I said I wouldn't want to merge this yet. If it turns you that it fixes the flakyness of our tests then we have to think about it...

kensipe:
and what flakiness?   I would expect flakiness to be a matter of test design.. not the kuttl itself... are you seeing something different?

ANeumann82:
Yes, we can hold that. I've created this branch to build a local kuttl version that I currently use to test the C* kuttl tests. I will run repeated TC tests over the weekend to test if this would fix the issue.

Wish it were the tests that were flaky, that would be easier to fix ;)

We're running basically into this issue:

```
failed to get file: command "docker exec --privileged kind-control-plane cat /kind/version" failed with error: exit status 1
```
https://github.com/kubernetes-sigs/kind/issues/1288

This is mostly triggered because for the failure recovery we create a kind cluster with 4 nodes which increases the chance of this happening.

ANeumann82:
So, as a quick update: All other flake causes are now fixed in the cassandra operator. There are still some flakes which come from this issue, but they are like 1 in 10 runs, so it's not too urgent to get this updated.

Still - at some point I think we need to push to the new version

ANeumann82:
@kensipe Any updates from your side? I'm not sure how we want to progress with this. I assume the issue you have with kind 0.8.1 is this one: https://github.com/kubernetes-sigs/kind/issues/1558

Can you use the mentioned workaround? 

Getting this into kuttl is not the highest priority, but at some point we do need to upgrade, i guess - and the issue in the older kind versions are responsible for flakes in Cassandra. Not that often, but it is kind of annoying.

pkg/test/step_test.go
```diff
@@ -287,8 +288,11 @@ func TestRun(t *testing.T) {
 					}),
 				},
 			}, func(t *testing.T, client client.Client) {
+				pod := testutils.NewPod("hello", testNamespace)
+				assert.Nil(t, client.Get(context.TODO(), types.NamespacedName{Namespace: testNamespace, Name: "hello"}, pod))
```
porridge, line 292: How is this related to the bump?

ANeumann82: I'm not fully sure, but if it's not in there the test fails on Line 295 with "object was modified". So fetching it once before fixes that.

ANeumann82:
> code gen scripts were off for some reason :(
> Aligned the versions with kudo project.
> we should merge [kudobuilder/kudo#1537](https://github.com/kudobuilder/kudo/pull/1537) around the same time

Yeah, I looked into that as well. I think it got adjusted because we had a custom code-generator to include a bugfix that is now finally merged and released. Thanks for helping out with that.

ANeumann82:
I looked into it: This PR needs to be merged and released before we can update and merge kudobuilder/kudo#1537

PR #96, opened 2020-05-15
Make it possible to increase KIND log verbosity.

porridge:

- retain KIND nodes on cluster creation failures, so there is something to collect logs from,
- introduce a notion of verbosity in `kuttl` and plumb it through to KIND logging

This makes it possible to debug KIND cluster failures such as the one described in https://github.com/mesosphere/kudo-cassandra-operator/pull/110#issue-418424225


zen-dog:
Does this give you more logs than already existing in the collected `artifacts/kind-*` bundle?

zen-dog:
On the other hand, I just had a situation where this additional logging helped a lot so 👍

porridge:
@zen-dog ,
> Does this give you more logs than already existing in the collected artifacts/kind-* bundle?

Yes, this is complementary:
- the `retain` means there will actually be something to collect in case cluster create fails mid-way,
- the logging means we will get output from `kind` itself

PR #99, opened 2020-05-18
Add `--timeout` Flag

kensipe:

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -146,7 +146,9 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 				options.ArtifactsDir = artifactsDir
 			}
 
-			options.Timeout = timeout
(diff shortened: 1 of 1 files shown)


pkg/kuttlctl/cmd/test.go
```diff
@@ -145,6 +146,8 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 				options.ArtifactsDir = artifactsDir
 			}
 
+			options.Timeout = timeout
```
jbarrick-mesosphere, line 149:
```suggestion
			if isSet(flags, "timeout") {
						options.Timeout = timeout
			}
```

jbarrick-mesosphere: This way the config file only gets overridden if the user specifies `--timeout`, otherwise it would always get set to the CLI's default.

kensipe:
haha I debated that.. and for some reason didn't go that way... you are totally right

PR #102, opened 2020-05-19
KEP04: TestStep.Apply Feature

kensipe:
TestStep has an `apply` property added which is a list of files.
As stated in KEP04, this will allow the reuse of files in multiple test steps.  In addition we have already seen a lot of `command: kubectl apply -f xyz.yaml` to enable this same behavior.   This provides a more concise way of expressing this common need.

Example:  `00-create.yaml`

```
apiVersion: kudo.dev/v1beta1
kind: TestStep
apply:
  - hello.yaml
  - hello2/hello2.yaml
  - ../hello3.yaml
  - directory
```
All the options above have been tested.  The relative path is relative to the step folder the TestStep is in.

### DIFF
--- a/pkg/file/files_test.go
+++ b/pkg/file/files_test.go
@@ -8,13 +8,45 @@ import (
 
 func TestFromPath(t *testing.T) {
 
-	paths, err := FromPath("testdata/path")
-	assert.NoError(t, err)
-	assert.Equal(t, 2, len(paths))
-	assert.Equal(t, "testdata/path/test1.yaml", paths[0])
+	tests := []struct {
+		name     string
+		path     string
+		pattern  string
+		expected []string
+		wantErr  bool
+	}{
+		{
+			name:     `good path no extension`,
+			path:     "testdata/path",
+			pattern:  "",
+			expected: []string{"testdata/path/skip.txt", "testdata/path/test1.yaml", "testdata/path/test2.yaml"},
+			wantErr:  false,
+		},
+		{
+			name:     `good path yaml extension`,
+			path:     "testdata/path",
+			pattern:  "*.yaml",
+			expected: []string{"testdata/path/test1.yaml", "testdata/path/test2.yaml"},
+			wantErr:  false,
+		},
+		{
+			name:     `bad path`,
+			path:     "testdata/badpath",
+			pattern:  "",
+			expected: []string{},
+			wantErr:  true,
+		},
+	}
 
-	_, err = FromPath("testdata/badpath")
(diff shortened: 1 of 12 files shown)


pkg/file/files_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestFromPath(t *testing.T) {
+
+	paths, err := FromPath("testdata/path")
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(paths))
+	assert.Equal(t, "testdata/path/test1.yaml", paths[0])
+
+	_, err = FromPath("testdata/badpath")
+	assert.Error(t, err, "file mode issue with stat testdata/badpath: no such file or directory")
```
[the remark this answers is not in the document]
kensipe: I find tables significantly harder to grok... they can be very useful but I wouldn't want to make things harder to read just to be consistent with tables

pkg/test/test_data/teststep-apply/00-assert.yaml
```diff
@@ -0,0 +1,14 @@
+# Verify that the pods were created.
+apiVersion: v1
+kind: Pod
+metadata:
+  name: hello
+status:
+  phase: Pending
```
zmalik, line 7: why verify `Pending` phase? why not just remove `status`? 
my question is this is because of current test setup that it stays in `Pending` or that's the desired assertion?

kensipe: it is the desired assertion for tests in the mocked control plane (which is also consistent with the other tests in this test suite.

jbarrick-mesosphere: Pods will never come out of pending in the local control plane mode since there's no kube-controller-manager running.

zmalik: but if we decide to run those tests in a kubernetes cluster we will need to change tests?
my question was if asserting `Pending` status we are verifying a mocked control-plane behavior? when the intent I see from test assertion is to verify that object is created

kensipe: tests do depend on knowing what env it will be run against... there are conditions you will never get to in a mock control plane... the expectation should be that tests against a mock control plane should be for an initial state... and tests against a cluster are end states (or controlled states).  It is too easy to "miss" a state.  beginning or intermediate states could be completely missed in a cluster resulting in flaky behavior.

PR #104, opened 2020-05-19
Fail on 2 TestSteps in 1 Step

### LINKED ISSUE #101: Limit TestSteps in Step to 1 or Merge
Currently if multiple TestSteps are provided for a Step 1 overwrites the other 
https://github.com/kudobuilder/kuttl/blob/master/pkg/test/step.go#L463

there is no log, and likely unexpected behavior..   Since a step can be defined across multiple files, it isn't as obvious as being in the same file.

We need to:
1. print a warning log message (we need to express which file teststep we are using and which files we are ignoring) or
2. return err or 
3. merge steps (less interesting as merge conflicts are impossible to automate)

leaning towards returning error

kensipe:
If there are 2 TestSteps defined in the same step (same file or across multiple files in the same step) this is an error condition now

Fixes #101


pkg/test/step_integration_test.go
```diff
@@ -341,3 +341,16 @@ func TestCheckedTypeAssertions(t *testing.T) {
 		})
 	}
 }
+
+func TestTwoTestStepping(t *testing.T) {
+	apply := []runtime.Object{}
+	step := &Step{
+		Name:            "twostepping",
```
jbarrick-mesosphere, line 348: :D

kensipe:
I'm not sure repeating all tests in multiple ways will add value... it will increase test time and the result is the same.   the different between reading the yaml and what is here is basically reading and marshalling the yaml  to a runtime.Object.  

do you see it differently?

porridge:
Your test case seems to be along the lines of:

> given a situation, where we already have parsed a Step, and we encounter another step in the file, raise an error

I guess I'm looking at it from a different angle, since the case I would write would be:

> given a file with two steps, raise an error

They do test the same code in the end, but I'd argue the latter is more cohesive - because the inputs are on the same higher level, while in your case one is an unparsed YAML input and the other is a Go data structure. This means easier to understand and more resilient to implementation changes.

Hope this makes sense.

ANeumann82:
I agree with @porridge here. I would prefer to have another file with two yaml test steps, and not this mix of "define one step in go code and load the other step via yaml", it makes it hard to understand the test.

kensipe:
working on it... however I don't see how it is hard to understand... with a precondition that a step already exists... if I read another it errors.   This is exactly what I would expect in a test... establish a precondition then put a subject under test.

PR #116, opened 2020-05-22
Commands can Opt-out of Logging

kensipe:
Provides the ability for a command to opt-out of logging.

This is useful when:
1. log may have sensitive information
2. log is too noisy due to a specific command

Co-authored-by: Marcin Owsiany <mowsiany@D2iQ.com>

Fixes #

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -114,6 +114,8 @@ type Command struct {
 	Background bool `json:"background"`
 	// Override the TestSuite timeout for this command (in seconds).
 	Timeout int `json:"timeout"`
+	// If set, the command is NOT logged.  Useful to sensitive logs or to reduce noise.
+	IgnoreLog bool `json:"ignorelog"`
 }
 
 // DefaultKINDContext defines the default kind context to use.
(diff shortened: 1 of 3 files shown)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -114,6 +114,8 @@ type Command struct {
 	Background bool `json:"background"`
 	// Override the TestSuite timeout for this command (in seconds).
 	Timeout int `json:"timeout"`
+	// If set, the command is NOT logged.  Useful to sensitive logs or to reduce noise.
```
porridge, line 117:
```suggestion
	// If set, the OUTPUT from the command is NOT logged.  Useful to sensitive logs or to reduce noise.
```

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -114,6 +114,8 @@ type Command struct {
...
+	IgnoreLog bool `json:"ignorelog"`
```
porridge, line 118:
```suggestion
	SkipOutputLog bool `json:"skipOutputLog"`
```

porridge: Not sure whether this should be `skipLogOutput` or `skipOutputLog`...

kensipe:
thanks @porridge !   I struggled with the naming a bit as well... and prefer your suggestions.  

thanks @ANeumann82 .  If debugging is needed they can always remove the flag.  Your comment gave me pause to consider `skipStdout` and `skipStderr`.   It seems like an extra burden to me... if you have a noising output... it shouldn't be hard to turn it off.

Issue #50, opened 2020-04-02
Add "script" setting to commands that runs as a bash script


jbarrick-mesosphere:
We should add a new setting to the commands struct called `script` that accepts a bash script as a string that we run. When running the script, we can set the `$NAMESPACE` and `$KUBECONFIG` variables. This will allow users to more easily write scripts inline when necessary.

```
apiVersion: kudo.dev/v1alpha1
kind: TestStep
commands:
- script: until ! kubectl get pod --namespace $NAMESPACE my-pod; do sleep 1; done
```

jbarrick-mesosphere:
*note* the example there can be achieved using `kubectl wait --for=delete pod --namespace $NAMESPACE my-pod`

kensipe:
I have the following questions:
1. currently Commands []Command `json:"commands"` is an array of Command I don't know how to union a type in an array (is that possible)...
2. we could keep it of type Command and add a "Script string " to the Command struct.. but that feels slightly awkward.. we to do if a command and a script is supplied?  we could some validation to error on it... or ignore one... but then I'm not sure if Background makes senses.  
3. What do you think of replacing script: for a specific script shell... I think I prefer to impl bash: instread... if we go that route perhaps we have a separate collections of scripts: like:

```
scripts:
  - bash: until ! kubectl get pod --namespace $NAMESPACE my-pod; do sleep 1; done
```

thoughts?

kensipe:
the unfortunate part of a separate collection is ordering... you may want to run a command before a script or vise versa 🤔

PR #117, opened 2020-05-22
Command Scripts from TestSuite or TestStep

kensipe:
Previously scripts added to `command:` would fail such as:
```
commands:
  - command: for i in {1..5}; do echo $NAMESPACE; done
  - command: sh -c "for i in {1..5}; do echo $NAMESPACE; done"
```
Both of these options fail... to provide some ability to script a `script:` is provided as a commands option such as:

```
commands:
  - script: for i in {1..5}; do echo $NAMESPACE; done
``` 

This works for TestSuite and TestStep

Fixes: #50 

Co-authored-by: Marcin Owsiany <mowsiany@D2iQ.com>
Co-authored-by: Andreas Neumann <aneumann@mesosphere.com>

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -108,6 +108,8 @@ type Command struct {
 	Command string `json:"command"`
 	// If set, the `--namespace` flag will be appended to the command with the namespace to use.
 	Namespaced bool `json:"namespaced"`
+	// Ability to run a script from TestStep (without a script file)
+	Script string `json:"script"`
 	// If set, exit failures (`exec.ExitError`) will be ignored. `exec.Error` are NOT ignored.
 	IgnoreFailure bool `json:"ignoreFailure"`
 	// If set, the command is run in the background.
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -108,7 +108,9 @@ type Command struct {
 	Command string `json:"command"`
 	// If set, the `--namespace` flag will be appended to the command with the namespace to use.
 	Namespaced bool `json:"namespaced"`
-	// Ability to run a script from TestStep (without a script file)
+	// Ability to run a shell script from TestStep (without a script file)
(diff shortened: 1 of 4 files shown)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -108,6 +108,8 @@ type Command struct {
 	Command string `json:"command"`
 	// If set, the `--namespace` flag will be appended to the command with the namespace to use.
 	Namespaced bool `json:"namespaced"`
+	// Ability to run a script from TestStep (without a script file)
```
porridge, line 111:
```suggestion
	// Ability to run a shell script from TestStep (without a script file).
```

pkg/test/utils/kubernetes.go
```diff
@@ -938,6 +938,17 @@ func ExpandEnv(c string, env map[string]string) string {
 func GetArgs(ctx context.Context, cmd harness.Command, namespace string, env map[string]string) (*exec.Cmd, error) {
 	argSlice := []string{}
 
+	if cmd.Command != "" && cmd.Script != "" {
+		return nil, errors.New("command and script can not be set in the same configuration")
+	}
+	if cmd.Command == "" && cmd.Script == "" {
+		return nil, errors.New("command or script must be configured")
```
porridge, line 945:
```suggestion
		return nil, errors.New("command or script must be set")
```

porridge: ?

pkg/test/utils/kubernetes.go
```diff
@@ -938,6 +938,17 @@ func ExpandEnv(c string, env map[string]string) string {
...
+	}
+
```
ANeumann82, line 947:
```suggestion
if cmd.Script != "" && cmd.Namespaced {
      return nil, errors.New("script can not used namespaced, use the NAMESPACE environment variable instead"
}
```

I think we should make sure that users don't try to combine these

kensipe: I'm beginning to think `namespaced` is a bad idea... we have better ways of working with namespace now... when we need it.   I'll add this in.  thanks!

PR #120, opened 2020-05-25
Bump Kubernetes dependencies to 1.18.3

nfnt:
This updated the Kubernetes APIs to 1.18.3. This update contains some breaking changes, as many calls have been made async ready (by taking a context.Context parameter) and consistent (by having options everywhere).


go.mod
```diff
@@ -29,14 +28,12 @@ require (
...
+	k8s.io/apiextensions-apiserver v0.18.3
+	k8s.io/apimachinery v0.18.3
+	k8s.io/client-go v0.18.3
+	k8s.io/code-generator v0.18.3
+	sigs.k8s.io/controller-runtime v0.6.0
+	sigs.k8s.io/controller-tools v0.3.0
 	sigs.k8s.io/kind v0.7.0
 )
-
-replace k8s.io/code-generator v0.17.3 => github.com/kudobuilder/code-generator v0.17.4-beta.0.0.20200316162450-cc91a9201457
```
porridge, line 42: Yay, we can finally get rid of https://github.com/kudobuilder/code-generator !
Should probably wait with deleting it until all branches catch up with this commit.

hack/update_codegen.sh
```diff
@@ -10,15 +10,13 @@ set -o pipefail
...
 version_and_repo=$(go list -f '{{if .Replace}}{{.Replace.Version}} {{.Replace.Path}}{{else}}{{.Version}} {{.Path}}{{end}}' -m k8s.io/code-generator)
 VERSION="$(echo "${version_and_repo}" | cut -d' ' -f 1 | rev | cut -d"-" -f1 | rev)"
-CODEGEN_REPO="$(echo "${version_and_repo}" | cut -d' ' -f 2)"
 REPO_ROOT="${REPO_ROOT:-$(git rev-parse --show-toplevel)}"
 CODE_GEN_DIR="${REPO_ROOT}/hack/code-gen/$VERSION"
 
 if [[ -d ${CODE_GEN_DIR} ]]; then
     echo "Using cached code generator version: $VERSION"
 else
-    echo "Cloning code-generator from ${CODEGEN_REPO}"
```
porridge, line 20: Why get rid of this? Does it cease to work with the `replace` removed?

porridge:
Also, `integration-test` is unhappy :-/

hack/update_codegen.sh
nfnt: Cloning didn't work for me, because the URL provided in `go.mod` isn't the one of the repo that needs to be used. That's why reverted this back to using a hardcoded repo. Because the first line of `git clone` is "Cloning $repo", I also removed this `echo`.

porridge: Hm, right. I tried to see how we could make it work in both cases but I got lost. `go help modules` says:

> Module code layout
> For now, see https://research.swtch.com/vgo-module for information
> about how source code in version control systems is mapped to
> module file trees.

And that doc is about the prototype and does not show an easy way to transform the module `Path` into a clonable URL. So let's stick with what you have for now.

go.mod
porridge: And with a similar commit in `kudobuilder/kudo` I guess!

nfnt: Yes, take a look at https://github.com/kudobuilder/kudo/pull/1537

kensipe:
nice work!   we need to discuss strategy of k8s version support.  A bump up in version potentially and likely means a pop off of a supported version... or an expansion of versions supported.
the request to hold means nothing more than that...  the solution looks good to me

kensipe:
this is a duplicate of https://github.com/kudobuilder/kuttl/pull/95 which has merged

PR #122, opened 2020-05-28
KEP 6 Proposal: KUTTL Test Phase Definition and Lifecycle Hooks

kensipe:


jbarrick-mesosphere:
This sounds great!

keps/0006-lifecycle.md
```diff
@@ -0,0 +1,207 @@
...
+* **TestStep** - Is the smallest component of a test, which are a collection of indexed files which govern they order of evaluation.  There are 4 basic types of TestStep files; 1) an apply manifest file (used for test setup), 2) an assert file, 3) an error file (used to assert the absense of a resource) and 4) a TestStep file which can run commands, update timeouts and delete resources. 
+
+### Proposal: Component Lifecycle
+
+When a test harness is run (be default with no flags for single test runs), the test harness will analyze all configured TestSuites establishing a memory model for execution (without analyzing every test / testfile).  A change from the current KUTTL, KUTTL will run each TestSuite as a unit (in order to support additional lifecycle hooks detailed in the next section). The order of TestSuite execution will be determined by the order of configuation in the kuttl-test.yaml configuration file.  Each TestSuite will run the set of tests defined by that suite with no guarantees on order.  Each Test will have a series of test steps which will run in index order.
```
nfnt, line 70: Can you explain a bit more how "run each TestSuite as a unit" changes the existing behavior? I understand this in the context of the lifecycle hooks mentioned below, but am not sure how this behavior would be different from the existing one. Except, of course, that there aren't any lifecycle hooks right now. Or maybe just remove the "A change from the current KUTTL" part :)

keps/0006-lifecycle.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Proposal: Component Lifecycle hooks
+
+It is proposed that the following activities are useful before and after certain test phases:
+
+1. apply manifests
+2. CRD management (create/delete)
+3. delete resources
+4. run commands
```
nfnt, line 79: Let's be clear that these activities are run in a specific order. Also, isn't CRD management included in the "apply manifests" and "delete resources" steps? Why does it have to be a separate step?

keps/0006-lifecycle.md
```diff
@@ -0,0 +1,207 @@
...
+    ├── 00-assert.yaml
+    ├── 00-install.yaml
+    ├── 01-assert.yaml
+    └── 01-resize.yaml
+```
+
+With a test suite file which look like:
+
+```
+apiVersion: kudo.dev/v1alpha1
```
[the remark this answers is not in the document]
kensipe: good catch... `kuttl.dev` 👍

kensipe:
@porridge agree with TestSuite vs TestCollection comment... I had the same thoughts and opted for the need to support the older option.  As you point out it is likely a transitive  thru deprecation.
I think I like what you named `kuttl-collection.yaml` to just be `kuttl.yaml`.  It is also not required which should be added in.

Now that we had a review pass, I will read thru all input and add / update... 

thanks @nfnt as well for the input... great suggestions

keps/0006-lifecycle.md
```diff
@@ -0,0 +1,207 @@
...
+kind: TestSuite
+before: 
+  commands:
+    - command: - command: ./bin/kubectl-kudo install --skip-instance ./repository/zookeeper/operator/
+after:
+  delete:
+  - apiVersion: kudo.dev/v1beta1
+    kind: OperatorVersion
+    name: zookeeper-0.3.0
```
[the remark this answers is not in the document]
kensipe: agree!

keps/0006-lifecycle.md
kensipe: CRDs are treated different in kuttl only in that there is a wait for the crd to be realized.... this is not true for manifests... any "wait" for arbitrary manifests is up to the user for an assert.

kensipe:
update 1 is grammar and language updates, kudo.dev -> kuttl.dev for apiVersion and toc.

Expect the following on the next update:
1. Different types added (`TestCollection`)
2. Different file support `kuttl.yaml` + `kuttl-suite.yaml`
3. Details on existing `kuttl-test.yaml` support (likely add an alternative proposal)
4. Details on errors / failures in tests and suites and the affect on collection / harness
5. Details on before and after function order.  (we will have to pick an order and or new structure... perhaps look for an alternative proposal here)

keps/0006-lifecycle.md
```diff
@@ -0,0 +1,341 @@
...
+
+### Proposal: Component Types
+
+The `kuttl.yaml` defined above will contain a new type which will consist of test harness control, testing defaults and collection of test suites. Based on this new type, it is proposed to establish a `kuttl.dev/v1beta2` GVK which will define the deprecation cycle.
+
+```
+apiVersion: kuttl.dev/v1beta2
+kind: TestCollection
+
+beforeCollection:
```
kensipe, line 102: originally this was `before:` and for the context perhaps is sufficient.
I starting thinking that in the future we might as `beforeEach` and `before:` becomes less clear.
I could see value in an `afterEach` that does an assert after each testsuite (in the case of a test-collection) or after each test (in the case of a test-suite).

This KEP is getting large and I think holding of on adding more at this point makes sense... lets discuss if you disagree

kensipe: perhaps we can add it in the KEP... we don't have to implement the kep all at once.

PR #123, opened 2020-05-29
teststep apply, assert and errors can now take urls, files or folder

### LINKED ISSUE #114: Apply and Manifests to support URLs
TestSuite.Apply takes directories and files.. but it should also use urls

example found in the wild
```
apiVersion: kudo.dev/v1alpha1
kind: TestSuite
manifestDirs:
  - ./test/manifests/
commands:
  - command: kubectl apply -f https://mesosphere.github.io/kubeaddons/bundle.yaml
testDirs:
  - ./addons/traefik/1.7.x/
```

should be:

```
apiVersion: kudo.dev/v1alpha1
kind: TestSuite
manifestDirs:
  - ./test/manifests/
  - https://mesosphere.github.io/kubeaddons/bundle.yaml
testDirs:
  - ./addons/traefik/1.7.x/
```

kensipe:
TestStep Apply, Errors and Asserts can now have urls (in addition to folders and files)

```
apiVersion: kudo.dev/v1beta1
kind: TestStep
apply:
  - https://mesosphere.github.io/kubeaddons/bundle.yaml
```
Is completely legit now.

There was an interesting situation with the kubeaddons url as well.  It has a file headers of:

```
# DO NOT EDIT: this file is generated

---
apiVersion: apiextensions.k8s.io/v1beta1
```

The 3 lines prior to the separator creates an unstructured object (without an error) but isn't valid.  This need detection has been added to the code as well.

Fixes #114

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -2,3 +2,4 @@ apiVersion: kudo.dev/v1beta1
 kind: TestSuite
 parallel: 4
 timeout: 120
+startControlPlane: true
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -2,4 +2,3 @@ apiVersion: kudo.dev/v1beta1
 kind: TestSuite
 parallel: 4
 timeout: 120
-startControlPlane: true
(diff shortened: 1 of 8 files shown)


kensipe:
As a side note:  it would be great to do this with manifests as well.. however today it is called `manifestsDir`... I think it would be best to switch to the new structure in KEP06 before adding url as an option.  We should call it `manifests:`

PR #133, opened 2020-06-15
fix for in-cluster case

jmccormick2001:
Signed-off-by: Jeff McCormick <jemccorm@redhat.com>


kensipe:
Thanks @jmccormick2001 !

PR #136, opened 2020-06-16
Errors Command

### LINKED ISSUE #125: Add KUTTL Assert Command
as `k kuttl test` is the command for testing a suite... there are times a user may want to assert something to be true (or not true).  A simple example is a test failure for which a solution was applied and you want to confirm that the change was correct... you don't want to run a full testsuite... you just want to assert the same conditions.

I see the value in asserting in 3 type of ways:
1. `k kuttl assert <folder>` would find all "assert" and "errors" yaml files and assert them
2. `k kuttl assert <path/to/file>` the find would be required to be yaml but otherwise will be asserted to be true.
3. `k kuttl assert <url/to/file?` same as #2 above but from an URL

It might be useful to have a `k kuttl errors` command to mirror assert which would assert what is an error if true.

This command could be great to assert conditions for a cluster prior to certain kinds of tests like a mixed workload test.

kensipe:
Errors command added which provides the ability to assert the absence of a manifest file on a cluster.  This is convenient when manually modifying a cluster and looking to assert that it is valid, or when creating an errors file.

when successful
```
go run cmd/kubectl-kuttl/main.go error --timeout 1 test.yaml
error assert is valid
```

when there is an assert failure
```
go run cmd/kubectl-kuttl/main.go errors -n temp test.yaml --timeout 2
resource matched of kind: /v1, Kind=Pod
Error: error asserts not valid
exit status 255
```

Fixes #125

### DIFF
--- a/pkg/kuttlctl/cmd/errors.go
+++ b/pkg/kuttlctl/cmd/errors.go
@@ -10,7 +10,7 @@ import (
 
 var (
 	errorsExample = `  # Asserts "errors" against a $KUBECONFIG cluster the values defined in the assert file.
-  kubectl kuttl errors <path/to/errorsfile.yaml>`
+  kubectl kuttl errors <path/to/errorsfile.yaml> <path/to/errorsfile.yaml>...`
 )
 
 // newErrorsCmd returns a new initialized instance of the errors sub command
--- a/pkg/kuttlctl/cmd/errors.go
+++ b/pkg/kuttlctl/cmd/errors.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 
 	"github.com/spf13/cobra"
+
+	"github.com/kudobuilder/kuttl/pkg/test"
 )
 
 var (
@@ -26,7 +28,7 @@ func newErrorsCmd() *cobra.Command {
 				return errors.New("one file argument is required")
 
 			}
-			return nil
+			return test.Errors(namespace, timeout, args...)
 		},
 	}
 
--- a/pkg/kuttlctl/cmd/errors.go
+++ b/pkg/kuttlctl/cmd/errors.go
@@ -1,0 +1,37 @@
+package cmd
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	errorsExample = `  # Asserts "errors" against a $KUBECONFIG cluster the values defined in the assert file.
+  kubectl kuttl errors <path/to/errorsfile.yaml>`
+)
+
+// newErrorsCmd returns a new initialized instance of the errors sub command
+func newErrorsCmd() *cobra.Command {
+	timeout := 5
+	namespace := "default"
+
+	errorsCmd := &cobra.Command{
+		Use:     "errors",
+		Short:   "Asserts the declared errors state to NOT be true.",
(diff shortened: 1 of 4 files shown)


pkg/test/step.go
```diff
@@ -515,23 +515,24 @@ func (s *Step) LoadYAML(file string) error {
...
 		}
 		return apply, nil
 	}
 
 	// it's a directory or file
-	paths, err := kfile.FromPath(filepath.Join(s.Dir, path), "*.yaml")
+	paths, err := kfile.FromPath(filepath.Join(dir, path), "*.yaml")
 	if err != nil {
-		return nil, fmt.Errorf("step %q apply %w", s.Name, err)
+		return nil, err
```
porridge, line 531:
```suggestion
		return nil, fmt.Errorf("failed to find YAML files in %s: %w", filepath.Join(dir, path), err)
```
Often a bare `err` propagated several stack frames up without a comment becomes quite cryptic..

pkg/kuttlctl/cmd/errors.go
```diff
@@ -0,0 +1,39 @@
...
+	namespace := "default"
+
+	errorsCmd := &cobra.Command{
+		Use:     "errors",
+		Short:   "Asserts the declared errors state to NOT be true.",
+		Long:    `Asserts the declared errors state provided as an argument to not be true in the $KUBECONFIG cluster. Valid arguments are a YAML file, URL to a YAML file.`,
+		Example: errorsExample,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("one file argument is required")
```
porridge, line 28: Check is for one or more, but error message is about exactly one?

pkg/kuttlctl/cmd/errors.go
```diff
@@ -0,0 +1,39 @@
...
+	"errors"
+
+	"github.com/spf13/cobra"
+
+	"github.com/kudobuilder/kuttl/pkg/test"
+)
+
+var (
+	errorsExample = `  # Asserts "errors" against a $KUBECONFIG cluster the values defined in the assert file.
+  kubectl kuttl errors <path/to/errorsfile.yaml>`
```
porridge, line 13: Should this also accept multiple files for consistency with the `assert` command?

pkg/kuttlctl/cmd/errors.go
kensipe: the check is for exactly 0... and the error message seems consist with assert and accurate.  at least 1 file is required.

pkg/kuttlctl/cmd/errors.go
kensipe: it does... perhaps  you are looking for the example to show or reflect that... looking

pkg/kuttlctl/cmd/errors.go
porridge: Let me try rephrasing: the message saying `one is required` is in my mind equivalent to `exactly one is required`. While in reality as you say in another comment, more than one is also fine.

kensipe: makes sense... I'll update!

PR #137, opened 2020-06-16
Namespace flag allows tests with no ns creation

### LINKED ISSUE #6: Support --namespace flag
KUTTL currently always creates namespaces for each test step, but some users may want to run their tests in a specific namespace or target an existing deployment. This would also be useful for running `kudo test` in less privileged environments where it is unable to create namespaces.

It is possible today to target static namespaces by specifying them on any commands or manifests, but it would be useful to add a `--namespace` flag to kudo test that controls the namespace used.

For example, if `--namespace` is set on `kudo test`, instead of creating namespaces for each test step and using the created namespace as the default for each resource, `kudo test` will set the default namespace to the `--namespace` flag and use that instead.

kensipe:
`test --namespace` support provided
rules:
1. The default namespace is ""
2. If default namespace, then tests will create their own namespaces for each tests (previous behavior)
3. if the namespace is provided then no namespaces will be created or deleted and the tests will run in that namespace.

Simple test run with "temp" ns
```
go run cmd/kubectl-kuttl/main.go test --start-control-plane=false ~/projects/playground/e2e/ -n temp
=== RUN   kuttl
    kuttl: harness.go:357: starting setup
    kuttl: harness.go:228: running tests using configured kubeconfig.
    kuttl: harness.go:303: running tests
    kuttl: harness.go:69: going to run test suite with timeout of 120 seconds for each step
=== RUN   kuttl/harness
=== RUN   kuttl/harness/pod
    kuttl/harness/pod: logger.go:42: 13:27:56 | pod | Ignoring pod.yaml as it does not match file name regexp: ^(\d+)-([^.]+)(.yaml)?$
=== PAUSE kuttl/harness/pod
=== CONT  kuttl/harness/pod
    kuttl/harness/pod: logger.go:42: 13:27:56 | pod/0-install | starting test step 0-install
    kuttl/harness/pod: logger.go:42: 13:27:56 | pod/0-install | Pod:temp/static-web created
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod/0-install | test step completed 0-install
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | pod events from ns temp:
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | 2020-06-16 13:27:56 -0500 CDT	Normal	Scheduled	Successfully assigned temp/static-web to kind-control-plane
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | 2020-06-16 13:27:56 -0500 CDT	Normal	Pulling	Pulling image "nginx"
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | 2020-06-16 13:27:57 -0500 CDT	Normal	Pulled	Successfully pulled image "nginx"
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | 2020-06-16 13:27:57 -0500 CDT	Normal	Created	Created container web
    kuttl/harness/pod: logger.go:42: 13:27:58 | pod | 2020-06-16 13:27:57 -0500 CDT	Normal	Started	Started container web
    kuttl: harness.go:342: run tests finished
--- PASS: kuttl (3.00s)
        --- PASS: kuttl/harness/pod (2.38s)
PASS

```

Fixes #6

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -57,7 +57,8 @@ type TestSuite struct {
 	ReportFormat *report.Type
 	// Namespace defines the namespace to use for tests
 	// The value "" means to auto-generate tests namespaces, these namespaces will be created and removed for each test
-	// Any other value is the name of the namespace to use and it must be created prior to tests running
+	// Any other value is the name of the namespace to use.  This namespace will be created if it does not exist and will
+	// be removed it was created (unless --skipDelete is used).
 	Namespace string
 }
 
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -55,6 +55,10 @@ type TestSuite struct {
 
 	// ReportFormat determines test report format (JSON|XML|nil) nil == no report
 	ReportFormat *report.Type
+	// Namespace defines the namespace to use for tests
+	// The value "" means to auto-generate tests namespaces, these namespaces will be created and removed for each test
+	// Any other value is the name of the namespace to use and it must be created prior to tests running
+	Namespace string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -163,6 +163,9 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 			}
 
 			if isSet(flags, "namespace") {
+				if strings.TrimSpace(namespace) == "" {
+					return errors.New(`setting namespace explicitly to "" or empty string is not supported`)
+				}
 				options.Namespace = namespace
 			}
 
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -53,6 +53,7 @@ func newTestCmd() *cobra.Command {
 	mockControllerFile := ""
 	timeout := 30
 	reportFormat := ""
+	namespace := ""
 
 	options := harness.TestSuite{}
 
@@ -161,6 +162,10 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 				options.ArtifactsDir = artifactsDir
 			}
 
+			if isSet(flags, "namespace") {
+				options.Namespace = namespace
+			}
+
 			if isSet(flags, "timeout") {
 				options.Timeout = timeout
 			}
@@ -214,6 +219,8 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 	testCmd.Flags().IntVar(&parallel, "parallel", 8, "The maximum number of tests to run at once.")
 	testCmd.Flags().IntVar(&timeout, "timeout", 30, "The timeout to use as default for TestSuite configuration.")
 	testCmd.Flags().StringVar(&reportFormat, "report", "", "Specify JSON|XML for report.  Report location determined by --artfacts-dir.")
+	testCmd.Flags().StringVarP(&namespace, "namespace", "n", "", "Namespace to use for tests. Provided namespaces must exist prior to running tests.")
+
 	// This cannot be a global flag because pkg/test/utils.RunTests calls flag.Parse which barfs on unknown top-level flags.
 	// Putting it here at least does not advertise it on a level where using it is impossible.
 	test.SetFlags(testCmd.Flags())
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -45,6 +45,7 @@ type Case struct {
 func (t *Case) DeleteNamespace(namespace string) error {
 	// if system defined namespace is "" we create and delete, otherwise don't
 	if t.Namespace != "" {
+		t.Logger.Log("Skipping deletion of user-supplied namespace:", namespace)
 		return nil
 	}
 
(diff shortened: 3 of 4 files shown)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -55,6 +55,10 @@ type TestSuite struct {
 
 	// ReportFormat determines test report format (JSON|XML|nil) nil == no report
 	ReportFormat *report.Type
+	// Namespace defines the namespace to use for tests
+	// The value "" means to auto-generate tests namespaces, these namespaces will be created and removed for each test
+	// Any other value is the name of the namespace to use and it must be created prior to tests running
```
porridge, line 60:
```suggestion
	// Any other value is the name of the namespace to use and it must exist prior to running tests
```

pkg/kuttlctl/cmd/test.go
```diff
@@ -161,6 +162,10 @@ For more detailed documentation, visit: https://kudo.dev/docs/testing`,
 				options.ArtifactsDir = artifactsDir
 			}
 
+			if isSet(flags, "namespace") {
+				options.Namespace = namespace
```
porridge, line 166: An interesting corner case is what happens when user specifies `--namespace ''` (this can happen intentionally or due to a misspelled variable name in a script for example).
Perhaps we should validate that the flag value is non-empty if specified explicitly?

pkg/test/case.go
```diff
@@ -42,6 +43,11 @@ type Case struct {
 
 // DeleteNamespace deletes a namespace in Kubernetes after we are done using it.
 func (t *Case) DeleteNamespace(namespace string) error {
+	// if system defined namespace is "" we create and delete, otherwise don't
+	if t.Namespace != "" {
+		return nil
```
porridge, line 48:
```suggestion
// DeleteNamespace deletes the auto-created namespace in Kubernetes after we are done using it.
func (t *Case) DeleteNamespace(namespace string) error {
	if t.Namespace != "" {
		t.Loggger.Log("Skipping deletion of user-supplied namespace:", namespace)
		return nil
```

pkg/test/case.go
```diff
@@ -42,6 +43,11 @@ type Case struct {
 
 // DeleteNamespace deletes a namespace in Kubernetes after we are done using it.
 func (t *Case) DeleteNamespace(namespace string) error {
```
porridge, line 45: It is really confusing to a casual reader right now is that we have the `namespace` arg and `t.Namespace`. They match usually, but not when `t.Namespace` has the sentinel value. We document the design in comments, which is fragile.

I think we should aim for self-documenting code whose semantics is obvious. How about making `Namespace` a `struct` with `Name string` and `Autocreated bool`. In that case we can get rid of the `namespace` arg altogether.

porridge: Same for `CreateNamespace`

pkg/test/case.go
porridge: Same for `CreateNamespace`

kensipe: yeah... I like it!

pkg/test/case.go
kensipe: I think I like the struct suggestion.. but requires some thinking... My initial thought is I'm expecting some refactoring shortly and perhaps it should go into that work.  If we don't do it here for this PR... I will make sure to capture it in an issue (so it is not lost).   also...  as discussed in a side conversation... we decided to create the supplied ns if it doesn't exist, on the premise that if the ns doesn't exist, the test would fail.  The main reason to not create a ns, is the user running kuttl may not have the rights to create a ns.  If the tests were going to fail anyway... attempting to create the ns will fail for another reason but if they have rights has the chance of succeeding.   This would change the nature of these checks... at min everything is potentially "auto-created" reducing the value of the suggested struct.  However perhaps there is room for something similar...  I'm with you!  self-documenting code is better... also localizing decisions is better.

pkg/kuttlctl/cmd/test.go
kensipe: nice... agree!   --namespace="" explicit should mean cluster scope which we do not allow and should explicit check for... thanks @porridge

pkg/test/case.go
kensipe: @porridge turns out... this looks really good with the stucts.  thank you sir!

pkg/test/case.go
```diff
@@ -33,16 +34,27 @@ type Case struct {
...
 
 	Client          func(forceNew bool) (client.Client, error)
 	DiscoveryClient func() (discovery.DiscoveryInterface, error)
 
 	Logger testutils.Logger
 }
 
+type namespace struct {
+	Name        string
+	AutoCreated bool
```
porridge, line 47: The code is clearer now, I am still concerned that `Case struct` contains a `Namespace` field, and there is also an `ns` variable. The distinction between their purpose is not clear at first sight.

If you don't think we could eliminate one of these, maybe we could at least rename `Case.Namespace` to something line `Case.NamespaceSpec` to make it clearer that it is a recipe for discovering/generating the actual namespace to use?

pkg/test/case.go
```diff
@@ -165,7 +200,26 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
...
-	t.CollectEvents(ns)
+	t.CollectEvents(ns.Name)
+}
+
+func (t *Case) determineNamespace() (*namespace, error) {
+	ns := &namespace{
+		Name:        t.Namespace,
+		AutoCreated: false,
+	}
+	//ns := t.Namespace
```
porridge, line 211: Is this line a genuine comment or a remnant of debugging? If the former than I have to admit it's not very clear.

pkg/test/case.go
kensipe: it isn't a namespace "spec"... but I get your point... the issue to be considered is if we make `case.namespace` a `namespace` type, than do we make `TestSuite` of that type... I'm real hesitate at this point, because this is externally exposed in the test suite file.  Let me think on this one.

kensipe: hmm... after more thought I think I understand why your suggestion better on naming... I'm real hesitate to use the "spec" word as it carries different meaning in the k8s world.  still thinking...

PR #138, opened 2020-06-17
Move from Master to Main

kensipe:

Fixes https://github.com/kudobuilder/kudo/issues/1564


kensipe:
any use of "master" in the repo are links to repos outside our control

kensipe:
yeah @porridge I always thought of it as master copy... and like many things is in language can take on different means or potentially be a trigger word.   This change is an effort to be more inclusive (as much as possible) and aligns with others in the industry making the same effort.

PR #139, opened 2020-06-23
Enhanced In Cluster Detection which Fixes Command Runs Out of Cluster

kensipe:
PR https://github.com/kudobuilder/kuttl/pull/133 made things work for in-cluster runs of kuttl, however it broke out of cluster runs of commands.  This fixes that with a check for in-cluster.

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -227,7 +227,8 @@ func (h *Harness) Config() (*rest.Config, error) {
 	} else {
 		h.T.Log("running tests using configured kubeconfig.")
 		h.config, err = config.GetConfig()
-		if err == nil {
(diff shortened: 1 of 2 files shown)


pkg/test/harness.go
```diff
@@ -227,7 +227,8 @@ func (h *Harness) Config() (*rest.Config, error) {
 	} else {
 		h.T.Log("running tests using configured kubeconfig.")
 		h.config, err = config.GetConfig()
-		if err == nil {
+		inK, _ := testutils.InClusterConfig()
+		if err == nil && inK {
```
porridge, line 231:
```suggestion
		inCluster, _ := testutils.InClusterConfig()
		if err == nil && inCluster {
```

PR #140, opened 2020-06-23
Pod Log Collector for Assert Failures

### LINKED ISSUE #16: test step - pod log output via API
this is a feature request, for the kuttl API to provide a means to obtain the Pod log when a test step executes a Pod, the API would let you get back the Pod log via the API call to execute a Test Step.

Our use case includes running custom container images within a Test Step, those images will need to produce log output in a particular format (e.g. v1alpha2 in our case), then we can aggregate or do other processing on the test log output, our test logs from the pods will show various kinds of feedback useful for the end-user (e.g. suggestions or error descriptions).

kensipe:
Provides Pod Log output on Assert or Error failure.
This is configured in the `TestAssert` Object and can be done with a pod name or a label selector.

For instance adding the following to an assert file:
```
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - selector: app=nginx
```

or this:
```
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - pod: static-web     
```

A namespace can be provided or it will default to the current test namespace.

The "collectors" is a list and multiple collectors can be provided.  They provide an output only if there is an assertion error and just prior to the output of the errors.  An example looks like:

```
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | collecting log output for [label: app=nginx]
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | running command: [kubectl logs --prefix -l app=nginx -n kudo-test-skilled-caribou]
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] 10-listen-on-ipv6-by-default.sh: Getting the checksum of /etc/nginx/conf.d/default.conf
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] 10-listen-on-ipv6-by-default.sh: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
    kuttl/harness/pod: logger.go:42: 22:50:15 | pod/0-install | [pod/static-web/web] /docker-entrypoint.sh: Configuration complete; ready for start up

```

Fixes #16

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -148,7 +148,7 @@ type TestCollector struct {
 }
 
 func (tc *TestCollector) Command() *Command {
-	c := fmt.Sprintf("kubectl logs %s", tc.Pod)
+	c := fmt.Sprintf("kubectl logs --prefix %s", tc.Pod)
 	ns := tc.Namespace
 	if tc.Namespace == "" {
 		ns = "$NAMESPACE"
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -1,6 +1,8 @@
 package v1beta1
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -90,6 +92,8 @@ type TestStep struct {
 	// Commands to run prior at the beginning of the test step.
 	Commands []Command `json:"commands"`
 
+	// AssertionCollectors is a set of pod log collectors fired on an assert failure
+	AssertionCollectors []TestCollector `json:"assertionCollectors,omitempty"`
 	// Allowed environment labels
 	// Disallowed environment labels
 }
@@ -130,6 +134,33 @@ type Command struct {
 	Timeout int `json:"timeout"`
 	// If set, the output from the command is NOT logged.  Useful for sensitive logs or to reduce noise.
 	SkipLogOutput bool `json:"skipLogOutput"`
+}
+
+// TestCollector are post assert / error commands that allow for the collection of information sent to the test log
+type TestCollector struct {
+	//TODO (kensipe): switch to TestAssert file AND add labels
+	// The pod name to access logs.
+	Pod string `json:"pod"`
+	// namespace to use.
+	Namespace string `json:"namespace"`
+	// Container in pod to get logs from
+	Container string `json:"container"`
+}
+
+func (tc *TestCollector) Command() *Command {
+	c := fmt.Sprintf("kubectl logs %s", tc.Pod)
+	ns := tc.Namespace
+	if tc.Namespace == "" {
+		ns = "$NAMESPACE"
+	}
+	c = fmt.Sprintf("%s -n %s", c, ns)
+	if tc.Container != "" {
+		c = fmt.Sprintf("%s -c %s", c, tc.Container)
+	}
+	return &Command{
+		Command:       c,
+		IgnoreFailure: true,
+	}
 }
 
 // DefaultKINDContext defines the default kind context to use.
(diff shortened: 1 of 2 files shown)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -132,5 +137,58 @@ type Command struct {
...
+	}
+	if len(tc.Selector) > 0 {
+		fmt.Fprintf(&b, " -l %s", tc.Selector)
+	}
+	ns := tc.Namespace
+	if len(tc.Namespace) == 0 {
+		ns = "$NAMESPACE"
+	}
+	fmt.Fprintf(&b, " -n %s", ns)
+	if len(tc.Container) != 0 {
```
porridge, line 166: How about consistently using `> 0` or `!= 0`?

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -132,5 +137,58 @@ type Command struct {
 	SkipLogOutput bool `json:"skipLogOutput"`
 }
 
+// TestCollector are post assert / error commands that allow for the collection of information sent to the test log
+type TestCollector struct {
+	// The pod name to access logs.
+	Pod string `json:"pod"`
+	// namespace to use.
+	Namespace string `json:"namespace"`
+	// Container in pod to get logs from
```
porridge, line 146: Might be worth mentioning here that container name is required if the pod contains more than one of them (since the `kubectl logs` fails in that case).

Alternatively, I think it might be a nice usability feature if we instead collected all containers' (including init ones which might have been injected at runtime) logs by default. This seems easy to do these days by just appending `--all-containers`.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -132,5 +137,58 @@ type Command struct {
 	SkipLogOutput bool `json:"skipLogOutput"`
 }
 
+// TestCollector are post assert / error commands that allow for the collection of information sent to the test log
```
porridge, line 140:
```suggestion
// TestCollector are post assert / error commands that allow for the collection of information sent to the test log.
// At least one of `pod` or `selector` is required.
```

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -132,5 +137,58 @@ type Command struct {
...
+type TestCollector struct {
+	// The pod name to access logs.
+	Pod string `json:"pod"`
+	// namespace to use.
```
porridge, line 144:
```suggestion
	// Namespace to use. The current test namespace will be used by default.
```

pkg/apis/testharness/v1beta1/test_types.go
porridge, line 146:
```suggestion
	// Container in pod to get logs from.
```

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -132,5 +137,58 @@ type Command struct {
...
+	Namespace string `json:"namespace"`
+	// Container in pod to get logs from
+	Container string `json:"container"`
+	// Selector is a label query to select pod
```
porridge, line 148:
```suggestion
	// Selector is a label query to select pod.
```

pkg/apis/testharness/v1beta1/test_types.go
kensipe: @porridge agree!   `--all-containers` added if no container specified.

PR #141, opened 2020-06-23
Kind Cleanup

kensipe:
This PR provides some clean up around working with kind server

1. ignore the kind generated log dirs if they exist
2. provide a make target to clean the log dirs
3. provide better setup as we kept getting failures detailed below because the cluster wasn't ready.

There was a common failure in particular with a built-in kind server creation where kind was still materializing which resulted in:
```
 kuttl/harness/pod: case.go:161: pods "static-web" is forbidden: error looking up service account kudo-test-national-mammal/default: serviceaccount "default" not found
``` 
Added a `WaitForSA` when working with any cluster, which is NOT applied for the following:
1. in-cluster detection (sa would have to be present)
2. running the mock control plane where it will never be present.

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -239,9 +239,9 @@ func (h *Harness) Config() (*rest.Config, error) {
 
 	// if not the mocked control plane
 	if !h.TestSuite.StartControlPlane {
-		// newly started cluster aren't ready until default service account is ready
+		// newly started clusters aren't ready until default service account is ready
 		// fixes: error looking up service account <namespace>/default: serviceaccount "default" not found
-		// no need for this with "inCluster" as we are self evident of it's existence.
+		// we avoid this with "inCluster" as the cluster must be already be up since we're running on it
 		err = testutils.WaitForSA(h.config, "default", "default")
 		if err != nil {
 			return h.config, err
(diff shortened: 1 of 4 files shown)


pkg/test/harness.go
```diff
@@ -237,6 +237,17 @@ func (h *Harness) Config() (*rest.Config, error) {
 		return h.config, err
 	}
 
+	// if not the mocked control plane
+	if !h.TestSuite.StartControlPlane {
+		// newly started cluster aren't ready until default service account is ready
```
porridge, line 242:
```suggestion
		// newly started clusters aren't ready until default service account is ready
```

pkg/test/harness.go
```diff
@@ -237,6 +237,17 @@ func (h *Harness) Config() (*rest.Config, error) {
...
+		// fixes: error looking up service account <namespace>/default: serviceaccount "default" not found
+		// no need for this with "inCluster" as we are self evident of it's existence.
```
porridge, line 244: I think it is still possible for the `default/default` service account to be missing when we're running in-cluster (e.g. due to a misconfiguration while we are in another namespace) but it's better to fail straight away on this later rather than uselessly delay here.

```suggestion
		// we avoid this with "inCluster" as the cluster must be already be up since we're running on it
```

PR #144, opened 2020-06-24
Removing of Duplication Printing of Cmd

kensipe:


kensipe:
good thoughts @porridge !   My initial reaction was moving the log to the beginning of the func... but IMO there is value in logging after `GetArgs` as env command expansion happens there which seems valuable to log.   To your point... the better solution is likely to keep command context in errors.  On it!

PR #147, opened 2020-06-25
affixed to v5 but with minor and patch updates

kensipe:

Fixes #


kensipe:
to fix:

```
Deprecation: [@octokit/request-error] `error.code` is deprecated, use `error.status`.
```

PR #149, opened 2020-07-01
Test Dirs can be tarballs from URLs

kensipe:
It is now possible In the list of TestDirs to provide a URL to a tar or tgz file which is a suite of tests bundled.   This can be in the TestSuite file or on the command line.

From command-line: `kubectl kuttl test http://localhost:8000/e2e.tgz --start-control-plane=false`

Or from TestSuite

```yaml
apiVersion: kudo.dev/v1beta1
kind: TestSuite
parallel: 4
timeout: 120
testDirs:
  - /opt/test
  - http://localhost:8000/e2e.tar
```

This PR will be followed by the ability to provide a local tarball as well.

More context:
Prior to this work, I didn't realize that the kubeconfig file was created for kind was in a temp folder.  We are leveraging that folder for more now.  Any tarballs are downloaded to that folder.  This is specifically designed for any size tarball.  That tarball is then untarred in the temp folder and run from a folder like any of the other tests.


pkg/http/client.go
```diff
@@ -0,0 +1,136 @@
...
+// Download takes a url to download and a filepath to write it to
+// this will write the response to any url request to a file.
+func (c *Client) Download(url string, path string) error {
+	// Get the data
+	resp, err := c.Get(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
```
ANeumann82, line 74: I guess we should have a check if the file already exists here and error out if the file exists?

pkg/test/harness.go
```diff
@@ -356,6 +370,34 @@ func (h *Harness) RunTests() {
...
+	// preprocessing step
+	for _, dir := range h.TestSuite.TestDirs {
+		if http.IsURL(dir) {
+			err := h.initTempPath()
+			if err != nil {
+				h.T.Fatal(err)
+			}
+			client := http.NewClient()
+			h.T.Logf("downloading %s", dir)
+			filePath, err := client.DownloadFile(dir, h.tempPath)
```
ANeumann82, line 385: If I read the code correctly we may have a problem when we encounter sth like this:

```
TestDirs:
  - http://someHost/suiteA/tests.tgz
  - http://someHost/suiteB/tests.tgz
```

kensipe: oh yeah... I did have in mind to solve that problem.. thx!

kensipe:
Thanks @ANeumann82 great review and feedback!   All concerns addressed.

PR #166, opened 2020-07-14
Add report commands to execute commands after tests are finished

ANeumann82:

- Add report commands which are executed after the test steps are done, but before the cluster is stopped.

Fixes #

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -1067,8 +1067,6 @@ func RunCommands(logger Logger, namespace string, commands []harness.Command, wo
 	}
 
 	for _, cmd := range commands {
-		logger.Logf("running command: %q", cmd.Command)
-
 		bg, err := RunCommand(context.Background(), namespace, cmd, workdir, logger, logger, logger, timeout)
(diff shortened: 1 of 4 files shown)


pkg/test/utils/kubernetes.go
```diff
@@ -1057,8 +1067,6 @@ func RunCommands(logger Logger, namespace string, commands []harness.Command, wo
 	}
 
 	for _, cmd := range commands {
-		logger.Logf("running command: %q", cmd.Command)
```
kensipe, line 1060: why is this removed... this keeps coming and going... and was resolved with marcin a couple of weeks ago.  It likely we need to review those closed PRs.

ANeumann82:
Yeah, all good points - This PR isn't supposed to be merged soon (or even at all), it's more an exploratory development. I was trying to integrate `kudo diagnostics collect` into KUDO Cassandra and needed a hook to run that after the tests were finished. 

I agree, we should focus on life-cycle support, I've also noticed that these `reportCommands` that are executed after all tests are ran are not necessarily what I need. A postTest command for each test case would be more what I need

PR #179, opened 2020-07-24
updated golangci linter to the latest

kensipe:

Fixes #

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -240,7 +240,7 @@ func writeXMLReport(dir, name string, ts *Testsuites) error {
 		return err
 	}
 	xmlStr := string(xDoc)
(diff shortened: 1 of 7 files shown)


pkg/report/report.go
```diff
@@ -240,7 +240,7 @@ func writeXMLReport(dir, name string, ts *Testsuites) error {
 		return err
 	}
 	xmlStr := string(xDoc)
-	return ioutil.WriteFile(file, []byte(xmlStr), 0644)
```
porridge, line 243: I'm not sure this is a good idea:
- if we keep `0644` and the user wants to restrict the permissions not owner-accessible-only, they can do this with `umask(1)`
- if we change to `0600` but the user does want world-readable, they will need an explicit `chmod` somewhere, which is messy

kensipe:
thanks @porridge resolved

PR #184, opened 2020-07-28
Command and Event collectors

### LINKED ISSUE #174: Add a Command TestAssert Collector
Recently added is the ability to have a TestAssert Collector for pod logs... When configured it will get the configured pod logs and add it to the test log.   There is a new desire to be able to get information from a command to gather context for the assertion failure.

The request
```
Do you think it would make sense to add an onFailureCommand field (or something like that) to TestCase commands?
Some background:
A TestCase for an operator I'm working on has (among others) a command to kubectl kudo install the operator and an instance of it, and another command to run a script which:
1. polls on the instance .spec.planExecution.status to make sure it's COMPLETE , and
2. dump some debugging information (plan status and k8s events) in case it never gets to COMPLETE
I could get rid of (1) by using the --wait flag of kubectl kudo install , but I still need the debug stuff from (2) on failure.
```

1. commands could be used to get plan status (which is a kudo specific thing)
2. seems like there is a desire to get k8s events (perhaps an event collector)

This may mean rethinking collectors based on types:
1. pod log collector
2. events collector
3. command collector

others?

kensipe:
PR #140 added Pod Log Collectors for Assertion failures.

This PR adds Events and command collectors.

To configure an Command:

```
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - type: command
    command: foo
```

Configuration of events allows for naming a pod or specifying a namespace

```
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - type: event
---
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - type: event
    namespace: xyz
---
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
collectors:
  - type: event
    namespace: xyz
    pod: zyz
```

Fixes #174


pkg/apis/testharness/v1beta1/collector.go
```diff
@@ -0,0 +1,164 @@
...
+	return &Command{
+		Command:       b.String(),
+		IgnoreFailure: true,
+	}
+}
+
+// String provides defaults of the type of collector
+func (tc *TestCollector) String() string {
+	if !tc.Valid {
+		return fmt.Sprintf("[collector invalid based on %s]", tc.InvalidReason)
```
porridge, line 137:
```suggestion
		return fmt.Sprintf("[collector invalid: %s]", tc.InvalidReason)
```

pkg/apis/testharness/v1beta1/collector.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if !(tc.Type == pod || tc.Type == events || tc.Type == command) {
+		return "unexpected confirmation issue with collector"
```
porridge, line 140: Honestly I don't understand what this message means.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -140,59 +137,27 @@ type Command struct {
...
-	return b.String()
+	Selector string `json:"selector,omitempty"`
+	// Type is a collector type which is pod, command or events
+	// command is default type if command field is not empty
+	// misconfiguration will lead to warning message in the logs
+	Type string `json:"type,omitempty"`
+	// Cmd is a command to run for collection.  It requires an empty Type or Type=command
+	Cmd           string `json:"command,omitempty"`
+	Valid         bool
+	InvalidReason string
```
porridge, line 160: I don't think these two should be a part of the API. Might be convenient to the implementation, but it makes the API messy.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -140,59 +137,27 @@ type Command struct {
...
-		}
-		fmt.Fprintf(&b, "label: %s", tc.Selector)
-	}
-	b.WriteString("]")
-	return b.String()
+	Selector string `json:"selector,omitempty"`
+	// Type is a collector type which is pod, command or events
+	// command is default type if command field is not empty
+	// misconfiguration will lead to warning message in the logs
+	Type string `json:"type,omitempty"`
```
porridge, line 156: How about putting `Type` as the first field in the struct?

pkg/test/step.go
```diff
@@ -414,7 +414,12 @@ func (s *Step) Run(namespace string) []error {
 		return testErrors
 	}
 	for _, collector := range s.Assert.Collectors {
+		collector.Validate()
 		s.Logger.Logf("collecting log output for %s", collector.String())
+		if !collector.Valid || collector.Command() == nil {
+			s.Logger.Log("skipping invalid assertion collector")
```
porridge, line 420: Please make this a fatal error. I find it extremely frustrating to find that the error handling code does not work exactly when I need it most to debug an elusive flake that just occured.

pkg/apis/testharness/v1beta1/test_types.go
kensipe: this is challenging...  these are not configurable JSON serializable fields... but fields to hold state between validation and execution.   this is the consequence of using the configuration API for internal state, which I would prefer not to do... but to change that would be huge at this point.
The only other options is to re-exec the validation code for the Command function.  at this point, I will switch to that.

pkg/apis/testharness/v1beta1/test_types.go
kensipe: happy to do so... but I don't understand why... there is a no  value to the order of these properties

pkg/test/step.go
kensipe: I can't see that this would make sense...   This is after all the tests have run.. and if we are here at all... the test is already failed... collectors are only run if there is a failure.

kensipe: @porridge the frustration is understandable...   as I think more about it... there is nothing more we can do here.
1. the testsuite is already marked as failed
2. there will be an entry in the log above for the configuration.
3. that log will have a good lead on what the issue is... i.e., `command collector requires a command`

porridge: Oh, I didn't realize we validate the config *after* we act on it!
How about validating at the beginning, before running anything?

pkg/apis/testharness/v1beta1/test_types.go
porridge: Just makes it easier for a human to understand the `struct` if the important fields are first.

pkg/apis/testharness/v1beta1/test_types.go
porridge: > consequence of using the configuration API for internal state

Oh, if we do it in more places then :man_shrugging:  I guess :cry: ...

PR #189, opened 2020-07-28
Suppress Events with Flag

### LINKED ISSUE #156: flag to suppress events from namespace in console
I am using kuttl to install and helm chart and verify that deployment and its pods are successfully started. 
Currently when test completes, we collectEvents gathers all events from namespace and prints it out to log.
Since I there are multiple pods started as part of deployment and each of them can have 10s of events, the console becomes noisy and even if test is passing it seems lot of things went wrong due to `WARNING` events from k8 for liveness and rediness. 

It would be great if we can make this control via flag option. 

this is done by code here: https://github.com/kudobuilder/kuttl/blob/fca4096b3e0002a15b1c9250041e79f816f0ccc6/pkg/test/case.go#L125

```
// CollectEvents gathers all events from namespace and prints it out to log
func (t *Case) CollectEvents(namespace string) {
	cl, err := t.Client(false)
	if err != nil {
		t.Logger.Log("Failed to collect events for %s in ns %s: %v", t.Name, namespace, err)
		return
	}

	eventsList := &eventsbeta1.EventList{}
```

sample noisy event logs
```
kuttl/harness/service-bootstrap: logger.go:42: 00:24:07 | service-bootstrap | 2020-07-02 23:20:51 +0530 IST	Warning	Unhealthy	Readiness probe failed: dial tcp 10.1.3.91:9092: connect: connection refused
    kuttl/harness/service-bootstrap: logger.go:42: 00:24:07 | service-bootstrap | 2020-07-02 23:20:51 +0530 IST	Warning	Unhealthy	Readiness probe failed: Get http://10.1.3.89:8081/: dial tcp 10.1.3.89:8081: connect: connection refused
    kuttl/harness/service-bootstrap: logger.go:42: 00:24:07 | service-bootstrap | 2020-07-02 23:20:51 +0530 IST	Warning	Unhealthy	Readiness probe failed:
    kuttl/harness/service-bootstrap: logger.go:42: 00:24:07 | service-bootstrap | 2020-07-02 23:20:52 +0530 IST	Warning	Unhealthy	Liveness probe failed: Get http://10.1.3.89:8081/: dial tcp 10.1.3.89:8081: connect: connection refused
```

kensipe:
Details in the Issue reported below.  This starts the ability to suppress output with the ability to support events.
`--suppress-log=events` will suppress events.

Thinking the following:
1. there is likely more things to suppress (commands, etc)
2. there might be value in having a "context" passed throughout the tests... where things like this can be added to the context.  That requires more thought.

This was a user request and looks to solve that issue for the next release.

Fixes #156

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -204,7 +204,7 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 	}
 
 	if funk.Contains(t.Suppress, "events") {
-		t.Logger.Logf("skipping event logging")
+		t.Logger.Logf("skipping kubernetes event logging")
 	} else {
 		t.CollectEvents(ns.Name)
 	}
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	petname "github.com/dustinkirkland/golang-petname"
+	"github.com/thoas/go-funk"
 	corev1 "k8s.io/api/core/v1"
 	eventsbeta1 "k8s.io/api/events/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -40,6 +41,8 @@ type Case struct {
 	DiscoveryClient func() (discovery.DiscoveryInterface, error)
 
 	Logger testutils.Logger
+	// Suppress is used to suppress logs
+	Suppress []string
 }
 
 type namespace struct {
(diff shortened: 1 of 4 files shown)


pkg/test/case.go
```diff
@@ -200,7 +203,11 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 		}
 	}
 
-	t.CollectEvents(ns.Name)
+	if funk.Contains(t.Suppress, "events") {
+		t.Logger.Logf("skipping event logging")
```
ANeumann82, line 207:
```suggestion
		t.Logger.Logf("skipping kubernetes event logging")
```

pkg/test/case.go
```diff
@@ -200,7 +203,11 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 		}
 	}
 
-	t.CollectEvents(ns.Name)
+	if funk.Contains(t.Suppress, "events") {
```
ANeumann82, line 206: I think it might make sense to make a constant for "events" and other types of logs that can be suppressed later, then we have a list of possible values for the cmd parameter

kensipe: yeah... I has that initial thought  as well... on it!

PR #195, opened 2020-08-07
linter should not change global tools

kensipe:

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,12 @@ help: ## Show this help screen
 
 .PHONY: lint
 lint: ## Run golangci-lint
-ifneq (${GOLANGCI_LINT_VER}, "$(shell golangci-lint --version | cut -b 27-32)")
-	./hack/install-golangcilint.sh
+ifneq (${GOLANGCI_LINT_VER}, "$(shell ./bin/golangci-lint --version 2>/dev/null | cut -b 27-32)")
+	@echo "golangci-lint missing or not version '${GOLANGCI_LINT_VER}', downloading..."
+	curl -sSfL "https://raw.githubusercontent.com/golangci/golangci-lint/v${GOLANGCI_LINT_VER}/install.sh" | sh -s -- -b ./bin "v${GOLANGCI_LINT_VER}"
 endif
-	golangci-lint run
-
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -31,11 +31,12 @@ help: ## Show this help screen
 
 .PHONY: lint
 lint: ## Run golangci-lint
-ifneq (${GOLANGCI_LINT_VER}, "$(shell golangci-lint --version | cut -b 27-32)")
-	./hack/install-golangcilint.sh
+ifneq (${GOLANGCI_LINT_VER}, "$(shell ./bin/golangci-lint --version 2>/dev/null | cut -b 27-32)")
```
porridge, line 34: Are you sure that works and will continue despite variable version string length? E.g. after `v1.99`?
```
$ golangci-lint --version|cut -b 27-32
v1.21.
```
Perhaps we should change this to `awk {print $4}` or something? Or `mv ./bin/golangci-lint ./bin/golangci-lint-$(GOLANGCI_LINT_VER)` after installation and stop using --version altogether?

kensipe: @porridge some time recently it appears that `golangci-lint` changed how they report version.  They used to have a `v` in front, as in your example.   They do not any longer.   what I get is 
```
golangci-lint --version | cut -b 27-32
1.27.0
```
This cut works for the correct version of linter.   Additionally:
* the version using your example would fail and would be installed
* the install is now NOT in the the global GOBIN, making control of this version easier to reason about
* one assumes that they will not keep changing this... and even if they did certain changes would be problematic for `awk` as well.   So this configuration is good for the current and foreseeable version.

Thanks for the awareness on this... IS there any other concern or reason to reject?

porridge: Thanks for making sure that it works.

My point is that even if the upstream does not change [this format template](https://github.com/golangci/golangci-lint/blob/cb58d1f82e6669b24541d17905d6287a80d0803c/pkg/commands/root.go#L20), then at a high enough version number our current `cut -b` parsing scheme *will* break. Because the version length is variable.

`awk '{print $4}'` would be more stable in this regard. But we can certainly wait with chnaging the parsing to that point.

PR #199, opened 2020-08-11
Adding verify generate to circle lint task

### LINKED ISSUE #198: Add CI Check for Missed Generated Files
Fix CI so that issues identified in https://github.com/kudobuilder/kuttl/pull/196 are captured 

in short:  `make generate` should never leave the repo dirty in CI

kensipe:

This adds a `verify-generate.sh` script which runs `make generate` and ensures that the repo isn't "dirty" after.  This is to ensure that changes which require generation are catch in CI.  This PR adds that check to the circleCI target as well.

Fixes #198

### DIFF
--- a/hack/verify-generate.sh
+++ b/hack/verify-generate.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+#####
+#  Used to ensure that running `make generate` results in a non-dirty repository.
+#  Used to verify that make generate wasn't forgotten for changes that may require generation.
+#####
+
+set -o nounset
+set -o pipefail
+# intentionally not setting 'set -o errexit' because we want to print custom error messages
+
+# versions of tools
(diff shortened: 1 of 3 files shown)


hack/verify-generate.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Invoking 'make generate' ends with non-zero exit code."
+    exit 1
+fi
+
+git diff --exit-code --quiet
+RETVAL=$?
+
+if [[ ${RETVAL} != 0 ]]; then
+    echo "Running 'make generate' produces changes to the current git status. Maybe you forgot to check-in your updated generated files?"
+    echo "The current diff: $(git diff)"
```
porridge, line 28:
```suggestion
    echo "The current diff:"
    git diff
```
No need to make the shell load all that output and pass it as a giant argument to echo. Might go bad if it's large.

PR #205, opened 2020-08-20
Allow the use of absolute paths in RuntimeObjectsFromPath

### LINKED ISSUE #204: TestStep apply prepends test directory even with an absolute path
**What happened**:
When applying a resource from an absolute path, the test directory was prepended.

When specifying an absolute path, the path should be used as written.

```bash
wget -O /tmp/pod.yaml https://gist.githubusercontent.com/joejulian/e9e2fdbacbf9ca76f617ff81f8fdcf86/raw/e7b4c3a8d2e880700c3aefff70e0d690b0b9aa52/pod.yaml
```

**./foo/00-foo.yaml**
```yaml
apiVersion: kudo.dev/v1beta1
kind: TestStep
apply:
  - /tmp/pod.yaml
```

```
kubectl-kuttl test .
```

```
...
    harness.go:365: step "foo" apply path /tmp/pod.yaml: failed to find YAML files in /home/jjulian/foo/test/tmp/pod.yaml: file mode issue with stat /home/jjulian/foo/test/tmp/pod.yaml: no such file or directory
...
```

**Environment**:
- KUTTL version (use `kubectl kuttl version`): KUTTL Version: version.Info{GitVersion:"0.6.0", GitCommit:"7b0360a", BuildDate:"2020-08-10T18:22:27Z", GoVersion:"go1.14.6", Compiler:"gc", Platform:"linux/amd64"}

joejulian:
Allow the use of absolute paths in RuntimeObjectsFromPath. Without this an absolute path still has the test directory prepended.

Fixes #204

### DIFF
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -553,8 +553,7 @@ func RuntimeObjectsFromPath(path, dir string) ([]runtime.Object, error) {
 	var err error
 	var paths []string
 
-	// if it's an absolute path
-	if len(path) > 0 && path[0] == filepath.Separator {
+	if filepath.IsAbs(path) {
 		paths, err = kfile.FromPath(path, "*.yaml")
 		if err != nil {
 			return nil, fmt.Errorf("failed to find YAML files in %s: %w", path, err)
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -550,9 +550,20 @@ func RuntimeObjectsFromPath(path, dir string) ([]runtime.Object, error) {
 	}
 
 	// it's a directory or file
-	paths, err := kfile.FromPath(filepath.Join(dir, path), "*.yaml")
-	if err != nil {
-		return nil, fmt.Errorf("failed to find YAML files in %s: %w", filepath.Join(dir, path), err)
+	var err error
+	var paths []string
+
+	// if it's an absolute path
+	if len(path) > 0 && path[0] == filepath.Separator {
+		paths, err = kfile.FromPath(path, "*.yaml")
+		if err != nil {
+			return nil, fmt.Errorf("failed to find YAML files in %s: %w", path, err)
+		}
+	} else {
+		paths, err = kfile.FromPath(filepath.Join(dir, path), "*.yaml")
+		if err != nil {
+			return nil, fmt.Errorf("failed to find YAML files in %s: %w", filepath.Join(dir, path), err)
+		}
 	}
 	apply, err := kfile.ToRuntimeObjects(paths)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


joejulian:
Regardless of use case, it's expected behavior that when you specify an absolute path to something, you get the something at that absolute path. Why would `/tmp/foo.yaml` end up picking up `/home/jjulian/go/src/<etc, etc>/tmp/foo.yaml`?

kensipe:
agree with @jbarrick-mesosphere that we want to not encourage non-portable test
agree with @joejulian that if it is absolute... well it should be absolute

IMO the proposal conflates 2 activities which I would like to separate.  I pushed a solution which has a path selection then has 1 execution path for getting the glob of yamls from that path.  I will likely move the `cleanPath` to `kfile` as it seems generically file related.

jbarrick-mesosphere:
I would rather have kuttl throw an error here when an absolute path is used, than allowing an absolute path to be used. What is the actual use case here? Why do we need to use an absolute path in a test?

kensipe:
thanks @joejulian !

joejulian:
I suppose that's fine. I can just replace `apply:` with `commands:` running kubectl apply.

joejulian:
> What's the use-case here?

> @joejulian joejulian mentioned this pull request 3 days ago
> add kubernetes-base-addons test for traefik mesosphere/kubeaddons-tests#7

jbarrick-mesosphere:
Yeah, I can see the pull request and the linked ticket, but they don't specify why this is needed. Can you elaborate?

joejulian:
```
apiVersion: kudo.dev/v1beta1
kind: TestStep
commands:
  - command: ./metallb-config.sh
  - command: kubectl create clusterrolebinding --clusterrole=cluster-admin --serviceaccount=${NAMESPACE}:kuttlaccount kuttlcrb
apply:
  - ${KUBEADDONS_ABS_PATH}/addons/konvoyconfig/0.0.x/konvoyconfig-3.yaml
  - ${KUBEADDONS_ABS_PATH}/addons/cert-manager/0.10.x/cert-manager-6.yaml
  - ${KUBEADDONS_ABS_PATH}/addons/metallb/0.9.x/metallb-1.yaml
  - ${KUBEADDONS_ABS_PATH}/addons/traefik/1.7.x/traefik-22.yaml
```
Where the environment is set to the path of the kubeaddons repo. That repo is not part of the tests repo and the tests cannot (for legacy reasons) be part of the kubeaddons repos.

jbarrick-mesosphere:
@joejulian thank you for answering my question.

PR #211, opened 2020-09-16
Support V1 CRDs

### LINKED ISSUE #210: 'pkg/test/utils.WaitForCRDs' doesn't work with V1 CRDs
**What happened**:
Calling `WaitForCRDs` on a V1 CRD will return the error "the following passed object does not match ..." because a V1Beta1 CRD is always assumed.

`WaitForCRDs` can wait on V1 and V1Beta1 CRDs.

Call `WaitForCRDs` with a V1 CRD.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19.1
- KUTTL version (use `kubectl kuttl version`): 0.6.1
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nfnt:
This adds V1 CRD API support to the Kubernetes helper functions. The helper `Scheme` will include the V1 CRD scheme, the `FakeDiscoveryClient` has the V1 CRD type added by default, and `WaitForCRDs` can handle V1beta1 and V1 CRDs.

Fixes #210

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -27,7 +27,8 @@ import (
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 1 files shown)


jbarrick-mesosphere (approved):
Nice find!

PR #212, opened 2020-09-17
Bump golangci-lint to use the new version subcommand.

porridge:

The new release has a `version --short` subcommand that makes it possible to get rid of the somewhat hacky parsing.


nfnt (approved):
LGTM!

zen-dog (approved):
🚢

PR #220, opened 2020-10-07
Add a `tail` parameter to pod collector.

### LINKED ISSUE #216: Allow specifying number of lines for `pod` test collector
`kubectl logs` defaults to just 10 trailing lines when a selector is used:
```
      --tail=-1: Lines of recent log file to display. Defaults to -1 with no selector, showing all log lines otherwise
10, if a selector is provided.
```

Please add an optional `tail` field that lets use specify the number of lines (or -1 for all), since 10 is sometimes insufficient.

ref: https://github.com/kudobuilder/kuttl/blob/c5f944beaaa4ecf92c55898fbf35b6b7c634c148/pkg/apis/testharness/v1beta1/collector.go#L115-L116

porridge:

Ability to set the number of lines to collect.
Also minor improvements to contributor guide.

Fixes #216

### DIFF
--- a/pkg/apis/testharness/v1beta1/collector_test.go
+++ b/pkg/apis/testharness/v1beta1/collector_test.go
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"testing"
 
-	"gotest.tools/assert"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestTestCollector_String(t *testing.T) {
@@ -153,7 +153,7 @@ func TestPodCommand(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			cmd := podCommand(&tt.tc)
-			assert.DeepEqual(t, cmd.Command, tt.cmd) // DeepEqual to get nice context diff on failure
+			assert.Equal(t, cmd.Command, tt.cmd)
 		})
 	}
 }
--- a/pkg/apis/testharness/v1beta1/collector_test.go
+++ b/pkg/apis/testharness/v1beta1/collector_test.go
@@ -3,6 +3,8 @@ package v1beta1
 import (
 	"strings"
 	"testing"
+
+	"gotest.tools/assert"
 )
 
 func TestTestCollector_String(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/apis/testharness/v1beta1/collector_test.go
```diff
@@ -3,6 +3,8 @@ package v1beta1
 import (
 	"strings"
 	"testing"
+
+	"gotest.tools/assert"
```
kensipe, line 7: please switch to testify (which will remove the new dependency as well with a go mod verify)

kensipe (changes requested):
love it!  thanks!

the project has a dependency on testify... could you switch out the newly added dependency.. then we are good!

kensipe:
briefly looked into this... I see you are getting DeepEqual... is there a way to do this with our current tooling?

kensipe:
value of DeepEquals

```
=== RUN   TestPodCommand/pod_name_with_set_tail
    collector_test.go:156: assertion failed: 
        --- cmd.Command
        +++ tt.cmd
          strings.Join({
          	"kubectl logs --prefix foo -n $NAMESPACE --all-containers --tail=",
          	"4",
        - 	"2",
        + 	"3",
          }, "")
        
```

value of testify `.Equal`

```
=== RUN   TestPodCommand/pod_name_with_set_tail
    collector_test.go:157: 
        	Error Trace:	collector_test.go:157
        	Error:      	Not equal: 
        	            	expected: "kubectl logs --prefix foo -n $NAMESPACE --all-containers --tail=42"
        	            	actual  : "kubectl logs --prefix foo -n $NAMESPACE --all-containers --tail=43"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-kubectl logs --prefix foo -n $NAMESPACE --all-containers --tail=42
        	            	+kubectl logs --prefix foo -n $NAMESPACE --all-containers --tail=43
        	Test:       	TestPodCommand/pod_name_with_set_tail
--- FAIL: TestPodCommand (0.00s)

```

pkg/apis/testharness/v1beta1/collector_test.go
porridge: Awesome, works even better!

porridge:
@kensipe I switched to testify and mentioned it in the contributor guide along with a couple other minor improvements there, PTAL.

kensipe (approved):
fantastic!  nice work!

PR #223, opened 2020-10-09
Include precise test case start in report output

porridge:

- rename `TestSuite.start` to `TestSuite.Timestamp` so that it is reported according to schema which says this is a required field
- rename `TestCase.start` to `TestCase.Timestamp` to also get the time for each test case. This is not in the schema, but should not hurt
- move the call to `testing.T.Parallel()` a bit earlier, so that the time measurements are not affected by it blocking

TL;DR: see PR subject

Background:

I'm doing capacity planning of a growing test suite. The high-level
quesiton I need to answer is: what cluster size to use for a given level of
parallelism of kuttl test cases to make sure that they can all pass.

Caveat: some of our cases *will deadlock* with each other when insufficient
resources are present on the cluster. They should probably use some form of
gang scheduling to avoid this problem, but currently they don't and it's not
under our control. So back to capacity planning.

In order to do that, I need to discover resource usage for each test case
separately. I have the resource monitoring covered, and I can run with disabled
parallelism for the purpose of data collection, but in order to attribute given
usage to the name of the test case in progress, I need to be able to reliably
and rather precisely tell when a given test case starts and stops.


porridge:
Oh, and another alternative I thought about is changing the existing format, but screen real estate is too precious...

kensipe:
thoughts...

first... great thinking!
months ago I had some similar thoughts but I think we are looking at from different angles... yours is more data driven which I like a lot but also requires some history of testing to determine optimization.
What I was thinking was establishing at a TestSuite level required minimum expectations for that TestSuite to run.  So if it is known by the author of the tests that for the tests to run it requires 3 nodes... then the tests will be skipped for clusters of < 3 nodes...  Other resource requirements acting the same (GPU, total concurrent cpu or max cpu requirement (like I need to land on a 8 core box)).
Personally I think both approaches are worth consideration.

That aside, I really like the PR for timing.   I believe we already have timing information.  if you use the `--report` flag there are times for test collections, test suites and individual tests as shown here: https://github.com/kudobuilder/kuttl/pull/128

```
 <testsuites name="" tests="10" failures="1" time="16.655560703s">
   <testsuite tests="10" failures="1" time="11.770504224s" name="pkg/test/test_data/">
     <testcase classname="test_data" name="list-pods" time="67.649654ms" assertions="1"></testcase>
     <testcase classname="test_data" name="teststep-assert" time="182.35219ms" assertions="2"></testcase>
     <testcase classname="test_data" name="teststep-errors" time="217.915925ms" assertions="4"></testcase>
     <testcase classname="test_data" name="teststep-apply" time="292.566614ms" assertions="2"></testcase>
     <testcase classname="test_data" name="delete-in-step" time="681.225526ms" assertions="4"></testcase>
     <testcase classname="test_data" name="create-or-update" time="844.747083ms" assertions="7"></testcase>
     <testcase classname="test_data" name="with-overrides" time="1.138570835s" assertions="4"></testcase>
     <testcase classname="test_data" name="crd-in-step" time="1.781726862s" assertions="2"></testcase>
     <testcase classname="test_data" name="deprecate-test" time="1.799839758s" assertions="2"></testcase>
     <testcase classname="test_data" name="cli-test" time="11.769091322s" assertions="2">
       <failure message="failed in step 1-patch" type="">resource Pod:kudo-test-suited-tomcat/cli-test-pod: .metadata.labels.test: value mismatch, expected: false != actual: true</failure>
     </testcase>
   </testsuite>
 </testsuites>
```

I believe the testsuite time is a calculated time (vs measured)... I think it takes the earliest testcase time and the latest testcase end time.  Same with test collections (named `testsuites` with an `s` at the end in the output)

Lets talk about if this fits your need or if we need to add more... or if there is more value in the log file instead of a report file.

porridge:
My longed-term term goal goes along the lines of your thoughts, but even a bit further. If we could annotate test cases with their resource requirements, and also give the harness some idea about available resources, it would be possible to have the harness automatically compute the maximum parallelism level, and schedule cases such that total resource consumption at any given moment would always be below the available amount.

The auto-generated parallelism level could even potentially be variable, since the distribution of resource consumption is usually far from uniform. Here is the data from my suite:

Zrzut ekranu z 2020-10-12 06-46-18

This way, the parallelism would increase when the less resource-hungry tests are run, optimizing run time and utilization, minimizing waste.

But as I said, step zero is discovering the actual resource consumption so that the test creator has something to annotate the tests with :slightly_smiling_face: 

So back to reporting the data:

I *knew* there was support for reporting this, but could not find it! It's well hidden!
Hits from [[site:kuttl.dev report]](https://www.google.com/search?client=firefox-b-1-e&q=site%3Akuttl.dev+report) are not useful, and help is also silent:
```
$ kubectl kuttl help test|grep report
$ kubectl kuttl help |grep report
$
```

I took a look at the report schema, and it already seems to have *most* of what I need, is machine-readable and this way does not need to pollute the log!
The windyroad schema seems very strict, but I found [evidence](https://community.atlassian.com/t5/Bamboo-questions/JUnit-XML-schema/qaq-p/769965) that even some big players take it with a grain of salt. I'm going to change this PR to adjust the report output a little.

kensipe:
I'll get on adding reports to doc site and increasing exposure to CLI.  that's a great point.

regarding xml format...  there are 2 drivers there... 1.  it was specifically requested by a user (operator-sdk scorecard which uses it).  2. all CI tools I know of have a way of displaying this information... which is why there are converters to it from other languages... it is strange that junit xml == java unit tests xml output... however there are lots of languages that convert to the "java" report... which is why https://github.com/jstemmer/go-junit-report is popular.   i guess it is first rights.

While the go-junit-report is popular it is dependent on the `go test` output which kuttl does not maintain.

kensipe:
looking into this... perhaps there is an older version being used.  I get report on the latest:
```
 k kuttl help test | grep report
      --report string                 Specify JSON|XML for report.  Report location determined by --artifacts-dir.
```
I still plan on raising more awareness

kensipe:
https://github.com/kudobuilder/kuttl/pull/227

porridge:
> looking into this... perhaps there is an older version being used. I get report on the latest:
> 
> ```
>  k kuttl help test | grep report
>       --report string                 Specify JSON|XML for report.  Report location determined by --artifacts-dir.
> ```

You're right. Outside of the testing container I have `v0.4.0` which is the most recent in `krew` index :cry:

porridge:
I changed the implementation to include the timestamps in the report output rather than logs.
It's even easier to ingest for my purpose.
WDYT @kensipe ?

jbarrick-mesosphere (approved):

jbarrick-mesosphere:
I didn't know about the report feature either, that's great!

PR #229, opened 2020-10-15
Bump kind to 0.9.0

jbarrick-mesosphere:
Note: this introduces a breaking change in that kind v1alpha3
configurations are no longer supported and the user must use kind
v1alpha4 instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,5 +32,5 @@ require (
 	k8s.io/code-generator v0.19.2
 	sigs.k8s.io/controller-runtime v0.6.3
 	sigs.k8s.io/controller-tools v0.3.0
-	sigs.k8s.io/kind v0.8.1
+	sigs.k8s.io/kind v0.9.0
 )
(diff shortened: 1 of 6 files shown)


ANeumann82:
Do we get a nice error if a user uses an old v1alpha3 config?

kensipe:
@ANeumann82 I've already added the dedupe logic for 1.19 depreciation warnings if that is what you are asking about.

kensipe:
@ANeumann82 I wasn't reading the full context... my previous response is not applicable...   I'll test...

kensipe:
@ANeumann82 the way we read the yaml file.. we decode the yaml into the `kindConfig.Cluster{}` which is an `v1alpha4.Cluster` without a check of type... this has the effect of forcing to `alpha4`.  With the `decoder.SetStrict(true)` I expect that a decode mismatch will thrown and err which will stop the test.  I'm good with that.  we could have better messaging... but regardless need docs cover some of this.

We should stop this PR... but https://github.com/kudobuilder/kuttl/issues/231 should make this a better experience.

kensipe (approved):
/lgtm

PR #230, opened 2020-10-16
Namespace Creation and Deletion Adhere to timeout rules

### LINKED ISSUE #217: Namespace deletion takes around 2 hrs sometimes during kuttl e2e tests
KUTTL Version: version.Info{GitVersion:"0.6.1", GitCommit:"4908507", BuildDate:"2020-08-25T18:22:14Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

Hello all,

I am facing an issue while running kuttl e2e test suite.
During any test cases failure, the namespace deletion takes around 2 hrs as below:

12:55:23     logger.go:42: 20:19:13 | abc-test | Deleting namespace: kudo-test-live-alpaca
14:56:34 === CONT  kuttl/harness/tri-11-create-one-or-more-test
14:56:34     logger.go:42: 22:20:24 | tri-11-create-one-or-more-test/3-add-test | test step failed 3-add-test

kuttl-config.yaml values:
apiVersion: kudo.dev/v1alpha1
kind: TestSuite
testDirs:
 - ./tests/e2e
parallel: 6
crdDir: ./config/crd/bases
timeout: 300
startKIND: true
artifactsDir: kuttl-artifacts
kindNodeCache: true

Is there any solution for this?
Please let me know if i missed any information.

Thanks,
Sanjeev

kensipe:
This change forces timeout rules to apply to namespaces. 

Thoughts:
1. There is likely value in separate kinds of timeouts (or configuration for separate concerns).  
2. timeout means a timeout for each individual step... there isn't a timeout for a test (which someone might want).  It could be challenging to "calculate" the overall timeout of a test which is each step timeout + ns timeout 

This seems like a good compromise for now.  The current result is:
1. if a namespace can not be deleted in the timeout, the test will fail

Fixes #217

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strconv"
 	"testing"
+	"time"
 
 	petname "github.com/dustinkirkland/golang-petname"
 	"github.com/thoas/go-funk"
@@ -64,7 +65,14 @@ func (t *Case) DeleteNamespace(ns *namespace) error {
 		return err
 	}
 
-	return cl.Delete(context.TODO(), &corev1.Namespace{
+	ctx := context.Background()
+	if t.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(t.Timeout)*time.Second)
+		defer cancel()
+	}
+
+	return cl.Delete(ctx, &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: ns.Name,
 		},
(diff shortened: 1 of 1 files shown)


ANeumann82 (approved):
Nice, that looks good 👍

pkg/test/case.go
```diff
@@ -64,7 +65,14 @@ func (t *Case) DeleteNamespace(ns *namespace) error {
...
 
-	return cl.Delete(context.TODO(), &corev1.Namespace{
+	ctx := context.Background()
+	if t.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(t.Timeout)*time.Second)
+		defer cancel()
+	}
+
+	return cl.Delete(ctx, &corev1.Namespace{
```
nfnt, line 75: If the context is canceled, this will return an error, right?

nfnt (approved):
LGTM! Just to clarify, the methods called will return an error if their context is canceled?

pkg/test/case.go
kensipe: yes... that is correct

PR #232, opened 2020-10-16
Adding warning message for non-supported version of kind config file.

### LINKED ISSUE #231: Detect use of  Deprecated API (Kind) and Warning
https://github.com/kudobuilder/kuttl/pull/229

Bumps the use of kind to a version that removes support for v1alpha3.  We should check and warn for a better UX.

```
check that `cluster.APIVersion` is > than a value.
```

kensipe:
Recent bump in Kind version https://github.com/kudobuilder/kuttl/pull/229 which had a break change in supported versions (in that it removed a version) surfaced the need to communicate this.  This PR does that.

The warning message is:
```
harness.go:601: Warning: "kind.sigs.k8s.io/v1alpha3" in kind.yaml is not a supported version.
```

It is possible if the configuration structure of earlier versions are still supported that everything runs fine... for this reason it was deemed not an error but a warning.  If there is a structural error... the decode code will throw an error... however it is nice to let users know that they need to consider looking at a version update even if theirs works.

Fixes #231

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -136,7 +136,7 @@ func (h *Harness) RunKIND() (*rest.Config, error) {
 		if h.TestSuite.KINDConfig != "" {
 			h.T.Logf("Loading KIND config from %s", h.TestSuite.KINDConfig)
 			var err error
-			kindCfg, err = loadKindConfig(h.TestSuite.KINDConfig)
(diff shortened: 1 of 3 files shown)


ANeumann82 (approved):
lgtm! 👍

jbarrick-mesosphere:
thanks!

PR #237, opened 2020-10-21
Support CRD v1 creation

alebedev87:
Thanks to [PR to support CRD v1](https://github.com/kudobuilder/kuttl/pull/211) it's now possible to wait for CRD of apiVersion v1.

However it's still not possible to create v1 CRDs using `crdDir` in the configuration.

This PR adds the ability to create v1 CRDs, in combination with the waiting mentioned earlier the CRD v1 support should be complete.

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -488,6 +488,7 @@ func LoadYAMLFromFile(path string) ([]runtime.Object, error) {
 	return LoadYAML(path, opened)
 }
 
+// LoadYAML loads all objects from a reader
 func LoadYAML(path string, r io.Reader) ([]runtime.Object, error) {
 	yamlReader := yaml.NewYAMLReader(bufio.NewReader(r))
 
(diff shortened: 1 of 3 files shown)


alebedev87:
@nfnt I would be glad if you can have a look at this PR. Any remarks/suggestions are very welcome.

kensipe:
@alebedev87 thanks for the PR... really great stuff... I want to analyze the approach... but regardless... we definitely want to support this!

I would prefer to see some tests added...  (expect a response within a day or 2)

kensipe:
my initial reaction was preferring a different approach for the `if`/`else` statement... perhaps a switch case would look better there.

pkg/test/utils/kubernetes.go
```diff
@@ -929,11 +931,23 @@ func WaitForCRDs(dClient discovery.DiscoveryInterface, crds []runtime.Object) er
...
-			waitingFor = append(waitingFor, schema.GroupVersionKind{
-				Group:   crd.Object["spec"].(map[string]interface{})["group"].(string),
-				Version: crd.Object["spec"].(map[string]interface{})["version"].(string),
-				Kind:    crd.Object["spec"].(map[string]interface{})["names"].(map[string]interface{})["kind"].(string),
-			})
+			spec := crd.Object["spec"].(map[string]interface{})
+			if v1Kind {
+				for _, ver := range spec["versions"].([]interface{}) {
+					waitingFor = append(waitingFor, schema.GroupVersionKind{
+						Group:   spec["group"].(string),
```
kensipe, line 938: seems like there is some repeating code... that we could extract into a func for the appending `waitFor`
I think I would prefer not to have an `else`... it would look better as a switch methinks.

kensipe (changes requested):
This is great!   we definitely want to support this and it wasn't even on our todo list... so thank you!

Beyond the readability nit (and refactor request).  If we could fit a test in that makes sense... that would be awesome!   Regardless with a little refactor we will get this merged!

pkg/test/utils/kubernetes.go
nfnt: IMO, the if-clause is fine here, I can't see how another switch would work here.

nfnt (approved):
LGTM, thanks for this great addition!

pkg/test/utils/kubernetes.go
alebedev87: I was also surprised that we always end up in the unstructured case but that's the way CRDs are loaded from the manifest files, only `kuttl.dev` or `kudo.dev` groups are converted to structured objects, the others are left unchanged: [here](https://github.com/kudobuilder/kuttl/blob/main/pkg/test/utils/kubernetes.go#L405).

alebedev87: I'm implementing a small unit test for the change I did, so, there will be a new commit soon.

alebedev87: @kensipe : extracted the part of GVK (GroupVersionKind) retrieval into a separate function. Added a unit test for it, so, this part is now better covered than before :)

kensipe: this went a different way than I expected... need to wrap my head around it.
the code redundancy I was hoping to remove wasn't... it just moved.
It seems more readable to me to have 1 code path for appending crds... there really isn't a different between 1 and * to me.  There is no need to differentiate 2 code paths for that IMO.

I will get to this today... shortly and throw in some thoughts.

can't thank you enough for surfacing this and provide a working solution.

alebedev87: The code redundancy you are referring to is the insertion into the slice of the objects to be waited? If so, I don't quite see the redundancy here as it's a simple, standard `append` to a slice.

Talking about the CRDs, they have different fields for different versions. `v1` allows you to have multiple versions whereas `v1beta1` allows you only one. From where comes the difference in the definitions of the golang structs. So, we need to make a `switch` between the types. But then there is also the fact that the CRDs are passed as unstructured objects after being read from the manifest files - this introduces the 2 code paths in the unstructured case. This is a pity, I agree, but that's the way CRDs are loaded, I didn't want my change to go that deep to refactor the loading from files too.

alenkacz (approved):
This looks great, thank you 👏

kensipe (approved):
nice work @alebedev87 !  and thank you!

PR #238, opened 2020-10-22
Make it possible to attach control-plane logs to stdout

alenkacz:
This is useful when for debugging tests or test setup.

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -24,7 +24,10 @@ type TestSuite struct {
 	StartControlPlane bool `json:"startControlPlane"`
 	// ControlPlaneArgs defaults to APIServerDefaultArgs from controller-runtime pkg/internal/testing/integration/internal/apiserver.go
 	// this allows for control over the args, however these are not serialized from a TestSuite.yaml
-	ControlPlaneArgs []string
+	ControlPlaneArgs []string `json:"controlPlaneArgs"`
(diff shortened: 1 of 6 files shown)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -25,6 +25,9 @@ type TestSuite struct {
 	// ControlPlaneArgs defaults to APIServerDefaultArgs from controller-runtime pkg/internal/testing/integration/internal/apiserver.go
 	// this allows for control over the args, however these are not serialized from a TestSuite.yaml
 	ControlPlaneArgs []string
+	// AttachControlPlaneOutput if true, attaches control plane logs (api-server, etcd) into stdout. This is useful for debugging.
+	// defaults to false
+	AttachControlPlaneOutput bool
```
ANeumann82, line 30:
```suggestion
	AttachControlPlaneOutput bool `json:"attachControlPlaneOutput"`
```

This should have a json mapping, right? I'm not sure why the controlPlaneArgs above don't have it...

ANeumann82 (approved):
lgtm, apart from the missing json mapping

pkg/apis/testharness/v1beta1/test_types.go
alenkacz: thanks! I've fixed it and added it to the above one as well

kensipe (approved):
This looks great!   thank!

One small thing (not required for merge)... all configurations are configurable outside of the kuttl config file.  There should be a flag for this. 

The kuttl-test.yaml config file is a convenience file but shouldn't be required.

PR #239, opened 2020-10-22
Fix: Kube Config Panic

kensipe:
In discovering what to write for an issue... fixed it :) 


alenkacz (approved):
👏 hahaha a good one :D thanks for the fix

PR #240, opened 2020-10-26
Log when skipping an unexpected resource kind

alenkacz:
Right now, if something in crd folder has an unexpected kind, we just silently skip it. This logs it so it makes it easy to debug this stuff.

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -573,6 +573,11 @@ func InstallManifests(ctx context.Context, client client.Client, dClient discove
 
 		for _, obj := range objs {
 			if len(kinds) > 0 && !MatchesKind(obj, kinds...) {
+				var expectedKinds []string
+				for _, k := range kinds {
+					expectedKinds = append(expectedKinds, k.GetObjectKind().GroupVersionKind().String())
+				}
+				log.Printf("Skipping resource %s because it does not match expected kinds: %s", obj.GetObjectKind().GroupVersionKind().String(), strings.Join(expectedKinds, ","))
 				continue
 			}
 
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -571,12 +571,13 @@ func InstallManifests(ctx context.Context, client client.Client, dClient discove
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


nfnt (approved):
LGTM!

pkg/test/utils/kubernetes.go
```diff
@@ -573,6 +573,11 @@ func InstallManifests(ctx context.Context, client client.Client, dClient discove
 
 		for _, obj := range objs {
 			if len(kinds) > 0 && !MatchesKind(obj, kinds...) {
+				var expectedKinds []string
+				for _, k := range kinds {
```
kensipe, line 577: this `for` loop is in a `for` loop and does not need to be... it is only creating `expectedKinds` for the log message after the `if` match... lets move it prior to the outer loop.

Otherwise... super smart!  thanks!

kensipe (changes requested):
Seems better to not nest the loops here

kensipe (approved):
change made... lgtm

pkg/test/utils/kubernetes.go
alenkacz: yeah I did not want to allocate that array if all kinds matches, which should be majority of cases

kensipe: hmm... that is a good point.. that intention would benefit from a comment... I'll revert.

kensipe: well... it could be viewed a number ways... (can't guess how this function might be used in the future).. currently it is called most of the time with 0 or 1 kinds... and with the merge of the CRDv1 PR it will be 0 or 2 kinds.   It is also only called during setup... It seems more readable to pull it out of the loop... wdyt?

alenkacz: my preference would still be to have it in the loop, it's closer to the usage so it's actually better for me even from readability perspective.

but I don't feel strongly about neither so whatever works for you better.

kensipe: didn't feel strongly about it... I'll revert (to your PR:))...  committing on another's PR is cool... until it isn't :) 
it is clear what the intention is... I'll add a comment and revert to your code structure.

PR #243, opened 2020-10-27
Report 0 testsuites and failed connections

### LINKED ISSUE #242: Report Always When Asked to Report
Current if KUTTL has a connection failure, or rbac issue could result in no tests when  `--report xml` is specified no report is generated.  This causes an issue in scorecard: https://github.com/operator-framework/operator-sdk/issues/4096 
It is expected that a report is always generated and if there is an error for the testsuite, that is communicated, allowing for communication via automation. 

Considering this a bug (I know... debatable :)

kensipe:
This PR provides the following benefits (read #242 for full context)

* if `--report` is specified, always report.  This will now report even if there are 0 test suites.
* if there is a connection issue the test is "failed".  If `--report` is specified, this will now report with the fatal error information.
* small adjustment for ` (h *Harness) fatal` which was always called with `error` which it was refactor to use.

Fixes #242

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -102,8 +102,10 @@ type Testsuites struct {
 	Properties *Properties `xml:"properties" json:"properties,omitempty"`
 	// Testsuite is a collection of test suites.
 	Testsuite []*Testsuite `xml:"testsuite" json:"testsuite,omitempty"`
-
-	start time.Time
+	// Failure defines a failure in this Testsuites. Not part of the Junit XML report standard, however it is needed to
+	// communicate test infra failures, such as failed auth, or connection issues.
+	Failure *Failure `xml:"failure" json:"failure,omitempty"`
+	start   time.Time
 }
 
 // NewSuiteCollection returns the address of a newly created TestSuites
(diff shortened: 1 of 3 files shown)


pkg/test/case.go
```diff
@@ -196,6 +196,7 @@ func shortString(obj *corev1.ObjectReference) string {
 func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 	ns := t.determineNamespace()
 	if err := t.CreateNamespace(ns); err != nil {
+		tc.Failure = report.NewFailure(err.Error(), nil)
```
jbarrick-mesosphere, line 199: Should we use AddFailure here?

pkg/report/report.go
```diff
@@ -235,6 +234,13 @@ func (ts *Testsuites) NewSuite(name string) *Testsuite {
 	return suite
 }
 
+// AddFailure adds a failure to the TestSuites collection for startup failures in the test harness
+func (ts *Testsuites) AddFailure(message string) {
```
jbarrick-mesosphere, line 238: `SetFailure` might be better since it's not a list

jbarrick-mesosphere (approved):
A couple nits, but LGTM!

pkg/report/report.go
kensipe: nice! thx

pkg/test/case.go
kensipe: the `AddFailure` (renamed to `SetFailure`) is at the TestSuites (collection level).  This failure is at the testcase level and will report a failure for each test case if all namespaces can not be created).  It might be debatable if we need to report each one individual.. but if we want to fail the whole suite on 1 namespace creation issue we would need to report from this run... and we currently don't

jbarrick-mesosphere (approved):

jmccormick2001 (approved):

Issue #146, opened 2020-06-25
Support "shouldFail" for Command


gpaul:
**What would you like to be added**:
Add "shouldFail" to the `v1beta1.command` type.

**Why is this needed**:
I'd like to confirm that executing a certain command fails.

porridge:
A command can fail in many ways - with certain error codes or terminated by a number of signals. Perhaps we should think of an interface that supports this, like a `expectedExitCode` (default 0) that we could match against the exit code?
Although then asserting `any non-zero exit code` would be difficult.

kensipe:
I'll group this in the "theme" of commands asserts... which has a number of concerns beyond what @porridge mentions. (thanks Marcin!).  I will target this "theme" of activity for v0.7.0 which is an iteration away.   
@gpaul more clarity for your use-case based on marcins comments would be great.

In fact, if you could provide a mock configuration and what your expectations around that would be really great!

gpaul:
Maybe something like

```
apiVersion: kudo.dev/v1alpha1
kind: TestStep
commands:
- command: ../../bin/dispatch login github --user=test --password=test
# Test that duplicate login fails.
- command: ../../bin/dispatch login github --user=test --password=test
  expectedExitCode: 1
```
This is the problem I was trying to test, by the way.

PR #245, opened 2020-10-28
KEP-0007: Assertions for CLI commands

nfnt:
By asserting CLI commands, kuttl can support many additional test scenarios that examining the internal state of an operator.

### DIFF
--- a/keps/0007-command-asserts.md
+++ b/keps/0007-command-asserts.md
@@ -1,0 +1,103 @@
+---
+kep-number: 7
+short-desc: Assertions for CLI commands
+authors:
+  - "@nfnt"
+owners:
+  - "@nfnt"
+creation-date: 2020-10-28
+last-updated: 2020-10-28
+status: provisional
+---
+
+# Assertions for CLI commands
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [User Story 1](#user-story-1)
+    - [User Story 2](#user-story-2)
+  - [Implementation Details](#implementation-details)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+By asserting CLI commands, kuttl can support many additional test scenarios that examing the internal state of an operator.
+
+## Motivation
+
+Operators and custom resources can be used to deploy complex applications. For example, KUDO provides operators for Apache Cassandra and Apache Kafka. Deploying custom resources of these operators may not always result in changed Kubernetes resources but also in changes that aren't surfaced by Kubernetes resources but are internal to the operator. Asserting these changes only works through accessing service APIs or running commands and parsing their output. kuttl should provide tools to assert such output.
+
+### Goals
+
+- Assert that commands examining the internal state of an operator succeed or fail
+- Keep eventual consistency when asserting commands
+
+### Non-Goals
+
+- Introduce "flows" to determine when to run commands. I.e. to run commands once before or after asserting resources
+
+## Proposal
+
+Introduce commands as part of `TestAssert`. These commands are run at a regular interval like resource assertions until they either succeed or a timeout of the assert is reached. If any of the commands fail, the assert is considered to have failed. A command's return value determines if a command failed or not.
+
+As a start, complex commands that parse output can be described using shell environments. At a later stage we can implement additional tools to simplify common tasks. E.g. to search with a regular expression in command output.
+
+### User Stories
+
+#### User Story 1
+
+The JVM options of a KUDO operator can be tuned using parameters. On deployment, the JVM options are logged in the pods running the application provided by the operator. To assert that a parameter change resulted in changed JVM options we need to run `kubectl logs pod-name` and search the logs for a specific string once the operator has been deployed.
+
+#### User Story 2
+
+An operator can be configured to encrypt it's API. To assert that the API endpoints are indeed encrypted we can run a specific `curl` command and check that it returns successfully once the operator has been deployed.
+
+### Implementation Details
+
+Command assertions use most fields of the existing commands for `TestStep` and follow their behavior. The `ignoreFailure`, `background` and `timeout` fields are removed as they are not needed here.
+
+An example that parses the logs of a pod:
+
+```yaml
+apiVersion: kuttl.dev/v1beta1
+type: TestAssert
+commands:
+  - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
+    namespaced: true
+```
+
+with `parse-logs.sh`:
+
+```bash
+#!/bin/bash
+
+POD=$1
+SEARCH=$2
+# $3 is "--namespace"
+NAMESPACE=$4
+
+kubectl logs -n ${NAMESPACE} ${POD} | grep ${SEARCH}
+```
+
+### Risks and Mitigations
+
+Some commands might be expensive to run and it would be better to run them once instead of running them repeatedly. This can be worked around by running commands as part of a test step instead. Commands in a test step run once.
+
+There could be subtle race conditions when running commands that have to run after a `TestStep` has been applied. The eventual consistency of Kubernetes might result in a command running in an environment where the changes of the `TestStep` haven't been applied yet. E.g., consider a `TestStep` that changes a `Deployment` and a command assert that examines the log of a `Pod` run by the `Deployment`. The change of the `Deployment` will restart pods, but as the restart is not immediate, the command might be run with the old pod.
+As a workaround, users can create an assert that doesn't have corresponding `TestStep`. This will then get asserted after the prior assert succeeded.
+
+## Graduation Criteria
+
+The `commands` field described in the [Implementation Details](#implementation-details) runs and asserts command return values.
+
+## Implementation History
+
+- 2020-10-28 - Initial draft (@nfnt)
--- a/keps/0007-command-asserts.md
+++ b/keps/0007-command-asserts.md
@@ -29,16 +29,16 @@ status: provisional
 
 ## Summary
 
-By asserting CLI commands, kuttl can support many additional test scenarios that examing the internal state of an operator.
+By asserting CLI commands, we can check operator state that isn't exposed through Kubernetes resources. This allow kuttl to support more test scenarios.
 
 ## Motivation
 
-Operators and custom resources can be used to deploy complex applications. For example, KUDO provides operators for Apache Cassandra and Apache Kafka. Deploying custom resources of these operators may not always result in changed Kubernetes resources but also in changes that aren't surfaced by Kubernetes resources but are internal to the operator. Asserting these changes only works through accessing service APIs or running commands and parsing their output. kuttl should provide tools to assert such output.
+The state of an operator is not always fully covered by Kubernetes objects. Some state may be expressed in stdout, logs or other tooling. kuttl should allow users to assert changes to this state. The simplest way to do this is by running commands or scripts that take care of asserting this state. E.g., by parsing log output. Such a command assertion can then either succeed or fail.
 
 ### Goals
 
 - Assert that commands examining the internal state of an operator succeed or fail
-- Keep eventual consistency when asserting commands
+- Assert commands in the same way Kubernetes objects are asserted
 
 ### Non-Goals
 
@@ -46,7 +46,7 @@ Operators and custom resources can be used to deploy complex applications. For e
 
 ## Proposal
 
-Introduce commands as part of `TestAssert`. These commands are run at a regular interval like resource assertions until they either succeed or a timeout of the assert is reached. If any of the commands fail, the assert is considered to have failed. A command's return value determines if a command failed or not.
+Introduce commands as part of `TestAssert`. These commands are run at a regular interval like resource assertions until all of them either succeed or the assertion timeout is reached. If any of the commands fail, the assert is considered to have failed. A command's return value determines if a command failed or not. I.e., command assertions behave in the same way as the existing resource assertions.
 
 As a start, complex commands that parse output can be described using shell environments. At a later stage we can implement additional tools to simplify common tasks. E.g. to search with a regular expression in command output.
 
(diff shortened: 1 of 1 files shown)


keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+By asserting CLI commands, kuttl can support many additional test scenarios that examing the internal state of an operator.
+
+## Motivation
+
+Operators and custom resources can be used to deploy complex applications. For example, KUDO provides operators for Apache Cassandra and Apache Kafka. Deploying custom resources of these operators may not always result in changed Kubernetes resources but also in changes that aren't surfaced by Kubernetes resources but are internal to the operator. Asserting these changes only works through accessing service APIs or running commands and parsing their output. kuttl should provide tools to assert such output.
+
+### Goals
+
+- Assert that commands examining the internal state of an operator succeed or fail
+- Keep eventual consistency when asserting commands
```
ANeumann82, line 41: Can we specify what we mean with "eventual consistency" here? My understanding is that this means that all asserted things must match at some point in time - it's not enough that Assertion 1 matched at time T1 and fails at T2, and Assertion 2 fails at T1 and matches at T2. (In this case, both Assertions matched at *some* point in time, but never together.)

Is this what's meant by "eventual consistency" here?

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+
+The JVM options of a KUDO operator can be tuned using parameters. On deployment, the JVM options are logged in the pods running the application provided by the operator. To assert that a parameter change resulted in changed JVM options we need to run `kubectl logs pod-name` and search the logs for a specific string once the operator has been deployed.
+
+#### User Story 2
+
+An operator can be configured to encrypt it's API. To assert that the API endpoints are indeed encrypted we can run a specific `curl` command and check that it returns successfully once the operator has been deployed.
+
+### Implementation Details
+
+Command assertions use most fields of the existing commands for `TestStep` and follow their behavior. The `ignoreFailure`, `background` and `timeout` fields are removed as they are not needed here.
```
ANeumann82, line 65: I think we need at least the `timeout` field as well - We should certainly monitor the executed command to make sure it doesn't run forever, and if we do that we should have the `timeout`.

For the `background` - I'm not sure. Could be useful, especially for things like the `kubectl logs pod-name`, I'd might want to add the `-f` to follow the logs and don't fetch them repeatedly if I know that a certain line will show up at some point. For this a background task might be good, as it wouldn't stop all other commands.

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+
+An example that parses the logs of a pod:
+
+```yaml
+apiVersion: kuttl.dev/v1beta1
+type: TestAssert
+commands:
+  - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
+    namespaced: true
+```
```
ANeumann82, line 75:
```suggestion
```yaml
apiVersion: kuttl.dev/v1beta1
type: TestAssert
commands:
  - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
    namespaced: true
timeout: 60
delay: 10
```
```
I would suggest an addition of an "assertDelay". This is currently fixed at 1 second, meaning the assert is checked after a delay of 1 second. 
When we add commands, we might want to increase this delay, maybe because of very expensive checks. 
An `initialDelay` might also make sense, although adding delays to check is often bad design...

File: keps/0007-command-asserts.md [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+
+#### User Story 1
+
+The JVM options of a KUDO operator can be tuned using parameters. On deployment, the JVM options are logged in the pods running the application provided by the operator. To assert that a parameter change resulted in changed JVM options we need to run `kubectl logs pod-name` and search the logs for a specific string once the operator has been deployed.
+
+#### User Story 2
+
+An operator can be configured to encrypt it's API. To assert that the API endpoints are indeed encrypted we can run a specific `curl` command and check that it returns successfully once the operator has been deployed.
+
+### Implementation Details
```
ANeumann82, line 63: I would like some details about how multiple commands would interact. 
- must they all succeed, or can they succeed one by one?
- are they executed in parallel or one after the other?

ANeumann82 (commented):
Nice work :) Some open questions though

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Motivation
+
+Operators and custom resources can be used to deploy complex applications. For example, KUDO provides operators for Apache Cassandra and Apache Kafka. Deploying custom resources of these operators may not always result in changed Kubernetes resources but also in changes that aren't surfaced by Kubernetes resources but are internal to the operator. Asserting these changes only works through accessing service APIs or running commands and parsing their output. kuttl should provide tools to assert such output.
+
+### Goals
+
+- Assert that commands examining the internal state of an operator succeed or fail
+- Keep eventual consistency when asserting commands
+
```
kensipe, line 42: we should define as a goal where in the life cycle a command assert will be invoked

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+### Non-Goals
+
+- Introduce "flows" to determine when to run commands. I.e. to run commands once before or after asserting resources
+
+## Proposal
+
+Introduce commands as part of `TestAssert`. These commands are run at a regular interval like resource assertions until they either succeed or a timeout of the assert is reached. If any of the commands fail, the assert is considered to have failed. A command's return value determines if a command failed or not.
+
+As a start, complex commands that parse output can be described using shell environments. At a later stage we can implement additional tools to simplify common tasks. E.g. to search with a regular expression in command output.
```
kensipe, line 51: the proposal should include:

* the difference from existing asserts (which is that they are fired until true and no more... other asserts will fire repeated until all asserts are true)
* we should define the output of a command assert failure (currently we have verbose diff on declarative asserts... this will clearly be different)
* we should define how the report output will be different and what it looks like
* we should define what defines a failure and what defines a success...  for example 0 == success and anything else is a failure?

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+### User Stories
+
+#### User Story 1
+
+The JVM options of a KUDO operator can be tuned using parameters. On deployment, the JVM options are logged in the pods running the application provided by the operator. To assert that a parameter change resulted in changed JVM options we need to run `kubectl logs pod-name` and search the logs for a specific string once the operator has been deployed.
+
+#### User Story 2
+
+An operator can be configured to encrypt it's API. To assert that the API endpoints are indeed encrypted we can run a specific `curl` command and check that it returns successfully once the operator has been deployed.
+
```
kensipe, line 62: it might be interesting to parse the test logs... I'm not sure how that would be done.. but we have the ability to have background processes which are used for in development operators... you may want to parse the controller output which is in the test stdout.

If not... we should add this to the list of future considerations at the bottom of this kep

keps/0007-command-asserts.md
kensipe: we need to be clear if commands are part of the timeout and how that applies
it also needs to be clear how to change the timeout for a specific command assert

(I'm thinking here... what if someone does a `tail -f` or `kubectl logs -f`... eventually this should timeout)

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+Command assertions use most fields of the existing commands for `TestStep` and follow their behavior. The `ignoreFailure`, `background` and `timeout` fields are removed as they are not needed here.
+
+An example that parses the logs of a pod:
+
+```yaml
+apiVersion: kuttl.dev/v1beta1
+type: TestAssert
+commands:
+  - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
```
kensipe, line 73: we should be clear if we are going to support all the configurations of the current commands... like `script`, etc.

we should provide all possible configurations in the examples or have several examples... or have defined struct

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+#!/bin/bash
+
+POD=$1
+SEARCH=$2
+# $3 is "--namespace"
+NAMESPACE=$4
+
+kubectl logs -n ${NAMESPACE} ${POD} | grep ${SEARCH}
+```
+
```
kensipe, line 89: I would be inclined to cycle through all cmd asserts first until they are successful... then run all declarative asserts repeatedly until complete.  (they all have to be true to succeed anyway)

Also... if someone wants to have a non-connected assert... they can always have 00-assert with declarative and 01-assert with commands.

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+### Risks and Mitigations
+
+Some commands might be expensive to run and it would be better to run them once instead of running them repeatedly. This can be worked around by running commands as part of a test step instead. Commands in a test step run once.
+
+There could be subtle race conditions when running commands that have to run after a `TestStep` has been applied. The eventual consistency of Kubernetes might result in a command running in an environment where the changes of the `TestStep` haven't been applied yet. E.g., consider a `TestStep` that changes a `Deployment` and a command assert that examines the log of a `Pod` run by the `Deployment`. The change of the `Deployment` will restart pods, but as the restart is not immediate, the command might be run with the old pod.
+As a workaround, users can create an assert that doesn't have corresponding `TestStep`. This will then get asserted after the prior assert succeeded.
```
kensipe, line 95: yes!   this is what I meant in my last comment... 00-assert with declaratives and 01-assert with commands.

however... we should consider the following:

```
apiVersion: kuttl.dev/v1beta1
type: TestAssert
commands:
  before:
    - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
  after:
    - command: ./parse-logs.sh name-of-pod-0 string-to-search-for
```

where we can find better words for `before` and `after`... and they mean before declarative asserts and after.

We could also just always have them after (opposite of previous earlier advice).   If they are always after we know that a transition can be confirmed before the command assert.  This is likely a common scenario IMO.

We should have these captured as alt proposals.   I'm leaning on the last recommendation now tho... cycling through all commands after declarative.

kensipe (commented):
Lets have a round of discussions and edits... love that this moving!  thank you!

keps/0007-command-asserts.md
kensipe: not a fan of delay... If we have one... we should be clear a delay from what... I would rather have the declarative asserts successful as a criteria of when to start commands... time delays are a common source of flaky tests :(

keps/0007-command-asserts.md
kensipe: +1 on parallel or not... that needs to be clear

keps/0007-command-asserts.md
nfnt: Yes, that's what's meant here. I'll clarify it in the KEP.

keps/0007-command-asserts.md
nfnt: A `TestAssert` already has a timeout parameter. Individual timeouts on commands create a challenge with multiple commands. If, say, command A has a timeout of 10 seconds, command B has a timeout of 30 seconds and command B, we check the command status every second... should we stop checking the status of command A after it has succeeded in the first 10 seconds but we're still waiting for command B to succeed? Having a single timeout as part of `TestAssert` makes the implementation much simpler as we don't have to support these edge cases.

keps/0007-command-asserts.md
nfnt: Good point. My initial though was to reuse what's already used by `TestStep` commands. It uses `testutils. RunCommands`, which runs the commands one by one. I'll document this.

keps/0007-command-asserts.md
nfnt: My idea is to cycle through the commands it the same loop that cycles through the objects. I'll detail this in the KEP. This would keep the implementation simple. Asserting commands before or after asserting Kubernetes objects could be added later.

keps/0007-command-asserts.md
```diff
@@ -0,0 +1,103 @@
...
+- [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+By asserting CLI commands, kuttl can support many additional test scenarios that examing the internal state of an operator.
+
+## Motivation
+
+Operators and custom resources can be used to deploy complex applications. For example, KUDO provides operators for Apache Cassandra and Apache Kafka. Deploying custom resources of these operators may not always result in changed Kubernetes resources but also in changes that aren't surfaced by Kubernetes resources but are internal to the operator. Asserting these changes only works through accessing service APIs or running commands and parsing their output. kuttl should provide tools to assert such output.
```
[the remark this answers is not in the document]
nfnt: Thanks for this! I'll change this section.

keps/0007-command-asserts.md
nfnt: I'll detail this in the implementation details.

keps/0007-command-asserts.md
nfnt: This is currently a non-goal, though it makes sense to add this later.

kensipe (commented):
worth considering "shouldFail" for commands... and lets add it to a goal or non-goal and future considerations if not a current goal.

https://github.com/kudobuilder/kuttl/issues/146

keps/0007-command-asserts.md
kensipe: @nfnt we likely need some clean up on timeouts perhaps it is just documentation.  There are inconsistencies with timeouts and how they are applied through kuttl.    if there is a TestAssert timeout (for ex. 30s).  Is it your expectation that 30s is the timeout for all of the Assert process... or for each command?

nfnt: Same behavior as before for `TestAssert`, this will be the loop:

```go
	for i := 0; i < timeout; i++ {
		// start fresh
		testErrors = []error{}
		for _, expected := range objects {
			testErrors = append(testErrors, s.CheckResource(expected, namespace)...)
		}

		for _, command := range commands {
			testErrors = append(testErrors, s.CheckCommand(command)...)
		}

		if len(testErrors) == 0 {
			break
		}

		time.Sleep(time.Second)
	}
```

As before, the timeout isn't a value in seconds, but number of iterations of this loop.

keps/0007-command-asserts.md
nfnt: The defined struct is here: https://github.com/kudobuilder/kuttl/pull/245/files#diff-c4ce7d6a93b893acac5e0afa3051a69f3c5944b8c1836f389a4424350f8a4989R68
There will be some renaming so that we can distinguish between a command used in `TestAssert` and a command used in `TestStep` because they have different fields.

keps/0007-command-asserts.md
ANeumann82: Well, we currently have a fixed delay of 1 second between rechecks. I.e. when an assert fails, there is a one-second delay until it is checked again. Might be worth to make that configurable.

ANeumann82 (approved):
I'm generally ok with the KEP - I still think it can be extended quite a bit, but that doesn't need to be done in the first implementation.

I'd like to see a couple of kens questions answered somwhere in the KEP, for example things about log output if a command fails, if the report output changes, and if so how.

keps/0007-command-asserts.md
nfnt: I added details around command output and exit codes to the proposal.

kensipe:
rebased to main

PR #247, opened 2020-10-28
Updating Linter: Adding gocritic

kensipe:
Updated Linters to be alphabetical... it was really hard to see or know if we used a specific linter
added in `gocritic`... and I like it!

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -77,7 +77,7 @@ var APIServerDefaultArgs = []string{
 	"--advertise-address={{ if .URL }}{{ .URL.Hostname }}{{ end }}",
 }
 
-//TODO (kensipe): need to consider options around AlwaysAdmin https://github.com/kudobuilder/kudo/pull/1420/files#r391449597
+// TODO (kensipe): need to consider options around AlwaysAdmin https://github.com/kudobuilder/kudo/pull/1420/files#r391449597
 
 // IsJSONSyntaxError returns true if the error is a JSON syntax error.
 func IsJSONSyntaxError(err error) bool {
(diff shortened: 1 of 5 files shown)


pkg/test/utils/kubernetes.go
```diff
@@ -395,13 +394,14 @@ func ConvertUnstructured(in runtime.Object) (runtime.Object, error) {
...
-	} else if (group == kudoGroup || group == kuttlGroup) && kind == "TestAssert" {
+	case (group == kudoGroup || group == kuttlGroup) && kind == "TestAssert":
 		converted = &harness.TestAssert{}
-	} else if (group == kudoGroup || group == kuttlGroup) && kind == "TestSuite" {
+	case (group == kudoGroup || group == kuttlGroup) && kind == "TestSuite":
 		converted = &harness.TestSuite{}
-	} else {
+	default:
 		return in, nil
 	}
```
ANeumann82, line 406: I wouldn't mind extracting the `(group == kudoGroup || group == kuttlGroup)` part into a separate "if" before the switch, so end up with a simpler switch:

```
if !(group == kudoGroup || group == kuttlGroup) {
  return in, nil
}
switch kind {
  case "TestStep:
    converted = &harness.TestStep{}
  case "TestAssert:
   converted = &harness.TestAssert{}
  case "TestSuite":
    converted = &harness.TestSuite{}
  default:
    return in, nil
}
```

ANeumann82 (approved):
Looks pretty good, I'd prefer to improve the one switch a bit, but not a blocker.

pkg/test/utils/kubernetes.go
kensipe: that makes a lot more sense!  thx

PR #250, opened 2020-11-03
Use WaitForCRDs from testenv

alenkacz:
Let's not reimplement what was already done in testenv and better.

This should not be released in bugfix version though because it potentially breaks API if someone was using that public method. Let's wait for minor release.


nfnt (approved):
Awesome! Yes, KUDO is using this method in some tests. Though using `envtest.WaitForCRDs` there should work as well.

pkg/test/harness.go
```diff
@@ -466,7 +466,10 @@ func (h *Harness) Setup() {
 		h.fatal(fmt.Errorf("fatal error installing crds: %v", err))
 	}
 
-	if err := testutils.WaitForCRDs(dClient, crds); err != nil {
+	if err := envtest.WaitForCRDs(h.config, crds, envtest.CRDInstallOptions{
+		PollInterval: 100 * time.Millisecond,
+		MaxTime:      10 * time.Second,
```
kensipe, line 471: I initially questioned the 10s max time... but then realized... that is what was there! :)
👏

kensipe (approved):
so much awesome!

kensipe:
this conflicts with the update merged to fix issue in 0.7.2
This should be a full replacement of that and takes precedence

kensipe:
removing hold... `main` is for next major release... if we need to push a 0.7.3 then we will create a release branch for it.

PR #251, opened 2020-11-03
Fix CRD handling in WaitForCRDs

alenkacz:
Current 0.7.1 build is broken and fails when installing CRDs. There was actually no test testing the CRD loading so I added it and then fixed it. The implementation is reused from controller runtime. We should merge this and release 0.7.2. Ultimately in 0.8 we should merge #250 and just drop all that code.

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -679,8 +679,10 @@ func NewPod(name, namespace string) runtime.Object {
 // NewCRDv1 creates a new CRD object of version v1
 func NewCRDv1(t *testing.T, name, group, resourceKind string, resourceVersions []string) runtime.Object {
 	crdVersions := []interface{}{}
+	served := true
 	for _, v := range resourceVersions {
-		crdVersions = append(crdVersions, interface{}(map[string]interface{}{"name": interface{}(v)}))
+		crdVersions = append(crdVersions, interface{}(map[string]interface{}{"name": interface{}(v), "served": interface{}(served)}))
+		served = false
 	}
 	return WithSpec(t, NewResource("apiextensions.k8s.io/v1", "CustomResourceDefinition", name, ""), map[string]interface{}{
 		"versions": crdVersions,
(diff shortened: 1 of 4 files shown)


pkg/test/utils/kubernetes.go
```diff
@@ -679,8 +679,10 @@ func NewPod(name, namespace string) runtime.Object {
 // NewCRDv1 creates a new CRD object of version v1
 func NewCRDv1(t *testing.T, name, group, resourceKind string, resourceVersions []string) runtime.Object {
 	crdVersions := []interface{}{}
+	served := true
 	for _, v := range resourceVersions {
-		crdVersions = append(crdVersions, interface{}(map[string]interface{}{"name": interface{}(v)}))
+		crdVersions = append(crdVersions, interface{}(map[string]interface{}{"name": interface{}(v), "served": interface{}(served)}))
+		served = false
```
alenkacz, line 685: I don't like the crd handling to be done like this but this code will ultimately go away in the next PR (see description)

pkg/test/utils/kubernetes_test.go
```diff
@@ -537,7 +532,7 @@ func TestExtractGVKFromCRD(t *testing.T) {
 					Kind:    "testresource3",
 				},
 			},
-			shouldError: true,
+			shouldError: false,
```
alenkacz, line 535: TBH I have no idea why this was true but anyway, it should not error in the current state

pkg/test/utils/kubernetes_test.go
```diff
@@ -599,11 +594,14 @@ func TestExtractGVKFromCRD(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			gotGVKs, err := ExtractGVKFromCRD(tt.inputCRDs)
 			if tt.shouldError {
-				assert.NotNil(t, err)
-				assert.Nil(t, gotGVKs)
+				if err == nil {
+					t.Errorf("%s: expecting err to be not nil but got nil", tt.name)
```
alenkacz, line 598: printint out names is very important to be able to debug failures

nfnt (approved):
LGTM!

kensipe (approved):
/lgtm 

nice stuff Alena!

PR #257, opened 2020-12-16
Warning for testdir Override

kensipe:
Feedback from community is that overriding the kuttl-test.yaml config testDirs is something that needed to be in the stdout log for awareness.

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -193,7 +193,7 @@ For more detailed documentation, visit: https://kuttl.dev`,
 			}
 
 			if len(args) != 0 {
-				log.Println("kutt-test config testdirs is overridden with arg: ", args)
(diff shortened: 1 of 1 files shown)


pkg/kuttlctl/cmd/test.go
```diff
@@ -195,6 +193,7 @@ For more detailed documentation, visit: https://kuttl.dev`,
 			}
 
 			if len(args) != 0 {
+				log.Println("kutt-test config testdirs is overridden with arg: ", args)
```
alenkacz, line 196: won't this end with a not nice looking log? Should we maybe rather do `strings.Join(args, ",")` ?

alenkacz (approved):
LGTM apart from that nit

pkg/kuttlctl/cmd/test.go
kensipe: +1. thanks @alenkacz !

PR #258, opened 2020-12-16
pkg/test: allow complex "assert" and "errors" step file names

estroz:
This PR improves file name matching regexps by adding meta char escapes and non-capturing groups, and permitts assert- and error-type steps to have complex names, ex. "00-assert-test-pod.yaml"

Signed-off-by: Eric Stroczynski <estroczy@redhat.com>

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	petname "github.com/dustinkirkland/golang-petname"
-	"github.com/thoas/go-funk"
+	funk "github.com/thoas/go-funk"
 	corev1 "k8s.io/api/core/v1"
 	eventsbeta1 "k8s.io/api/events/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -26,7 +26,8 @@ import (
 	testutils "github.com/kudobuilder/kuttl/pkg/test/utils"
 )
 
-var testStepRegex = regexp.MustCompile(`^(\d+)-([^.]+)(.yaml)?$`)
+// testStepRegex contains one capturing group to determine the index of a step file.
+var testStepRegex = regexp.MustCompile(`^(\d+)-(?:[^\.]+)(?:\.yaml)?$`)
 
 // Case contains all of the test steps and the Kubernetes client and other global configuration
 // for a test.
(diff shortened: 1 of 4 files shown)


kensipe:
nice!  thank you sir!   I like it.   I will review in more detail shortly... but a quick pass looks good!
(shortly being I will complete tonight)

kensipe (approved):
/lgtm

❤️  it!   thanks @estroz

pkg/test/case.go
```diff
@@ -304,6 +301,18 @@ func (t *Case) CollectTestStepFiles() (map[int64][]string, error) {
 	return testStepFiles, nil
 }
 
+// getIndexFromFile returns the index derived from fileName's prefix, ex. "01-foo.yaml" has index 1.
+func (t *Case) getIndexFromFile(fileName string) (int64, bool, error) {
+	matches := testStepRegex.FindStringSubmatch(fileName)
+	if len(matches) != 2 {
+		t.Logger.Log("Ignoring", fileName, "as it does not match file name regexp:", testStepRegex.String())
```
alenkacz, line 308: I would personally prefer to return error from here and log in the method that's handling the error

pkg/test/case.go
```diff
@@ -304,6 +301,18 @@ func (t *Case) CollectTestStepFiles() (map[int64][]string, error) {
...
+		return 0, false, nil
+	}
+
+	i, err := strconv.ParseInt(matches[1], 10, 32)
+	return i, true, err
```
alenkacz, line 313: is the hasIndex really necessary? I think just returning index and error should be fine here. Now having index is an error in my mind.

alenkacz (changes requested):
I am a bit lost here (not your fault, that's the state of the code as it was before you started working on it).

I feel like right now the file parsing logic is two places, but it's parsing the same files, right? The regexes are kind of similar but not equal 🤔 what am I missing here? Can we maybe make some cleanup in there since we're doing changes in that area already?

pkg/test/case.go
kensipe: I think alena means "Not having an index is an error"... "Now" -> "Not"

I could go either way...   I don't like to introduce errors for logic and this isn't an "error".   We allow for files that are not indexed in the test folders.   They are ignored.  The log message is to bring awareness to that fact.  

The err could be a parsing error in which case we exit `CollectTestStepFiles`
In the case of an ignored file... we continue looping the rest of the files.

kensipe: I think I would prefer a default index of -1  and remove the `hasIndex`.    thus it isn't an error and we can remove the extra bool... the check external to this would be 

```
if index == -1 {
    continue
}
```

kensipe (approved):
/lgtm

kensipe:
looking to see if @alenkacz is good... thanks Eric!

estroz:
@kensipe thanks!

@alenkacz the regexps are very similar and can probably be merged into one with a larger refactoring, however I don't think I have time to do one now. I will try when I have more time!

alenkacz (approved):
Thanks

PR #259, opened 2020-12-17
harness: add JSON tag for the `suppress` config option

estroz:
This PR adds JSON tags to fields missing them in `TestSuite`.

Signed-off-by: Eric Stroczynski <estroczy@redhat.com>


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -66,7 +66,7 @@ type TestSuite struct {
 	// be removed it was created (unless --skipDelete is used).
 	Namespace string
```
alenkacz, line 67: since you're already in here, would you mind adding the json mapping also for namespace? potentially all other things that does not have one?

alenkacz (approved):
LGTM. If you have time/energy it would be great to add the json mapping for other fields as well :)

kensipe (approved):
/lgtm

Issue #254, opened 2020-12-15
Multi-cluster test support


jbarrick-mesosphere:
A mechanism for testing across more than one cluster. Currently, it is only possible to apply manifests or asserts in one cluster for an entire suite, but in some cases it can be useful to have more than one.

My use-case for this is I am testing an application that is federated and need to verify state in the attached cluster.

To implement this, we could add an optional kubeconfig path to the `TestStep` object that allows overriding the kubeconfig used for the entire `TestStep`. To ensure flexibility, this setting could allow interpolating environment variables.

PR #261, opened 2021-01-08
Add KEP for multi-cluster support

jbarrick-mesosphere:
Adds a KEP describing how to implement multi-cluster tests.

#254

### DIFF
--- a/keps/0008-multi-cluster.yaml
+++ b/keps/0008-multi-cluster.yaml
@@ -1,0 +1,48 @@
+---
+kep-number: 8
+short-desc: This KEP describes how we will support more than one cluster under test
+title: KUTTL Mult-Cluster Support
+authors:
+  - "@jbarrick-mesosphere"
+owners:
+  - "@jbarrick-mesosphere"
+creation-date: 2021-01-08
+last-updated: 2021-01-08
+status: provisional
+---
+
+# KUTTL Multi-Cluster Support
+
+This KEP describes how we will support more than one cluster in KUTTL.
+
+## Table of Contents
+
+* [Summary](#summary)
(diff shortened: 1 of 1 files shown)


alenkacz (approved):
Looks good 👍

kensipe (approved):
this makes a lot of sense... the core concepts are there... questions:

1. is the intent to have the configured kubeconfig in a step to over ride for the entire step?  It seems so
2. If this is true, then a step is restricted to a cluster and to work with 2 cluster (apply) before an assert.. it would require 2 steps correct?
3. currently all kubectl and commands use the local kubeconfig file that is created at the start kuttl... I assume we don't overwrite this...  and I assume the intention is for a changed step that the change is true for commands and kubectl.   We will need to make sure to capture this.

this /lgtm
it would be great to provide answer in the doc itself to provide additional context to future kep reviewers.

nice work ::clap::

jbarrick-mesosphere:
> this makes a lot of sense... the core concepts are there... questions:
> 
>     1. is the intent to have the configured kubeconfig in a step to over ride for the entire step?  It seems so
> 
>     2. If this is true, then a step is restricted to a cluster and to work with 2 cluster (apply) before an assert.. it would require 2 steps correct?
> 
>     3. currently all kubectl and commands use the local kubeconfig file that is created at the start kuttl... I assume we don't overwrite this...  and I assume the intention is for a changed step that the change is true for commands and kubectl.   We will need to make sure to capture this.
> 
> 
> this /lgtm
> it would be great to provide answer in the doc itself to provide additional context to future kep reviewers.
> 
> nice work :clap:

Addressed these!

PR #272, opened 2021-02-04
KEP 07 Implementation

jlpedrosa:
This implements KEP 07 (Draft) by appending the AssertCommands to the TestStep structure.

```go
type TestAssertCommand struct {
	// The command and argument to run as a string.
	Command string `json:"command"`
	// If set, the `--namespace` flag will be appended to the command with the namespace to use.
	Namespaced bool `json:"namespaced"`
	// Ability to run a shell script from TestStep (without a script file)
	// namespaced and command should not be used with script.  namespaced is ignored and command is an error.
	// env expansion is depended upon the shell but ENV is passed to the runtime env.
	Script string `json:"script"`
	// If set, the output from the command is NOT logged.  Useful for sensitive logs or to reduce noise.
	SkipLogOutput bool `json:"skipLogOutput"`
}
```

There's no issue associated with KEP 07.

> Create TestAssertCommand struct, and add it as a list to TestAssert
> Create Helper functions in utils to run assertions'
> Update RunCommands to receive an existing context
> Update step Check method to also run the scripts/commands specified in the step assertion
> Update the set Run command to honor the timeout (as commands can be lengthy)

### DIFF
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"regexp"
@@ -355,12 +356,26 @@ func (s *Step) CheckResourceAbsent(expected runtime.Object, namespace string) er
 	return nil
 }
 
+// CheckAssertCommands checks if the commands defined in TestAsserts have been run successfully
+// the errors returned can be a a failure of executing the command or the failure of the command executed.
+func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, command []harness.TestAssertCommand, timeout int) []error {
+	testErrors := []error{}
+	if _, err := testutils.RunAssertCommands(ctx, s.Logger, namespace, command, "", timeout); err != nil {
+		testErrors = append(testErrors, err)
+	}
+	return testErrors
+}
+
 // Check checks if the resources defined in Asserts and Errors are in the correct state.
-func (s *Step) Check(namespace string) []error {
+func (s *Step) Check(namespace string, timeout int) []error {
 	testErrors := []error{}
 
 	for _, expected := range s.Asserts {
 		testErrors = append(testErrors, s.CheckResource(expected, namespace)...)
+	}
+
+	if s.Assert != nil {
+		testErrors = append(testErrors, s.CheckAssertCommands(context.TODO(), namespace, s.Assert.Commands, timeout)...)
 	}
 
 	for _, expected := range s.Errors {
@@ -391,7 +406,7 @@ func (s *Step) Run(namespace string) []error {
 				command.Background = false
 			}
 		}
-		if _, err := testutils.RunCommands(s.Logger, namespace, s.Step.Commands, s.Dir, s.Timeout); err != nil {
+		if _, err := testutils.RunCommands(context.TODO(), s.Logger, namespace, s.Step.Commands, s.Dir, s.Timeout); err != nil {
 			testErrors = append(testErrors, err)
 		}
 	}
@@ -402,13 +417,18 @@ func (s *Step) Run(namespace string) []error {
 		return testErrors
 	}
 
-	for i := 0; i < s.GetTimeout(); i++ {
-		testErrors = s.Check(namespace)
+	timeoutF := float64(s.GetTimeout())
+	start := time.Now()
+
+	for elapsed := 0.0; elapsed < timeoutF; elapsed = time.Since(start).Seconds() {
+		testErrors = s.Check(namespace, int(timeoutF-elapsed))
 
 		if len(testErrors) == 0 {
 			break
 		}
-
+		if hasTimeoutErr(testErrors) {
+			break
+		}
 		time.Sleep(time.Second)
 	}
 
@@ -444,7 +464,7 @@ func (s *Step) String() string {
 	return fmt.Sprintf("%d-%s", s.Index, s.Name)
 }
 
-// LoadYAMLFromFile loads the resources from a YAML file for a test step:
+// LoadYAML loads the resources from a YAML file for a test step:
 // * If the YAML file is called "assert", then it contains objects to
 //   add to the test step's list of assertions.
 // * If the YAML file is called "errors", then it contains objects that,
@@ -588,4 +608,13 @@ func cleanPath(path, dir string) string {
 		return path
 	}
 	return filepath.Join(dir, path)
+}
+
+func hasTimeoutErr(err []error) bool {
+	for i := range err {
+		if errors.Is(err[i], context.DeadlineExceeded) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -356,11 +356,11 @@ func (s *Step) CheckResourceAbsent(expected runtime.Object, namespace string) er
 	return nil
 }
 
-// CheckAssertCommands checks if the commands defined in TestAsserts have been run successfully
+// CheckAssertCommands Runs the commands provided in `commands` and check if have been run successfully.
 // the errors returned can be a a failure of executing the command or the failure of the command executed.
-func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, command []harness.TestAssertCommand, timeout int) []error {
+func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, commands []harness.TestAssertCommand, timeout int) []error {
 	testErrors := []error{}
-	if _, err := testutils.RunAssertCommands(ctx, s.Logger, namespace, command, "", timeout); err != nil {
+	if _, err := testutils.RunAssertCommands(ctx, s.Logger, namespace, commands, "", timeout); err != nil {
 		testErrors = append(testErrors, err)
 	}
 	return testErrors
--- a/pkg/test/step_integration_test.go
+++ b/pkg/test/step_integration_test.go
@@ -425,3 +425,100 @@ func TestStepFailure(t *testing.T) {
 	errs := step.Run(namespace)
 	assert.Equal(t, len(errs), 1)
 }
+
+func TestAssertCommandsValidCommandRunsOk(t *testing.T) {
+	
+	step := &Step{
+		Name:  			"commandRun",
+		Index: 			0,
+		Logger:          testutils.NewTestLogger(t, ""),
+		Client:          func(bool) (client.Client, error) { return testenv.Client, nil },
+		DiscoveryClient: func() (discovery.DiscoveryInterface, error) { return testenv.DiscoveryClient, nil },
+	}
+
+	// Load test that has an echo command, so it should run ok, and don't return any errors
+	err := step.LoadYAML("step_integration_test_data/assert_commands/valid_command/00-assert.yaml")
+	assert.NoError(t, err)
+
+	errors := step.Run("irrelevant")
+	assert.Equal(t, len(errors), 0)
+}
+
+func TestAssertCommandsMultipleCommandRunsOk(t *testing.T) {
+	
+	step := &Step{
+		Name:  			"commandRun",
+		Index: 			0,
+		Logger:          testutils.NewTestLogger(t, ""),
+		Client:          func(bool) (client.Client, error) { return testenv.Client, nil },
+		DiscoveryClient: func() (discovery.DiscoveryInterface, error) { return testenv.DiscoveryClient, nil },
+	}
+
+	// Load test that has an echo command, so it should run ok, and don't return any errors
+	err := step.LoadYAML("step_integration_test_data/assert_commands/multiple_commands/00-assert.yaml")
+	assert.NoError(t, err)
+
+	errors := step.Run("irrelevant")
+	assert.Equal(t, len(errors), 0)
+}
+
+func TestAssertCommandsMissingCommandFails(t *testing.T) {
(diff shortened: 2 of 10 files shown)


pkg/test/utils/kubernetes.go
```diff
@@ -1071,6 +1071,32 @@ func RunCommand(ctx context.Context, namespace string, cmd harness.Command, cwd
 	return nil, err
 }
 
+// toAssertCommands converts a set of TestAssertCommand to Commands so it all the existing functions can be used 
+// that expect Commands data type.
+func toAssertCommands(assertCommands []harness.TestAssertCommand, timeout int) (commands []harness.Command) {
```
kensipe, line 1076: This does not below in `kubernetes.go` if it goes in `utils` at all...  `utils` predates you (and me) and would prefer better packaging... don't let that block you... but it shouldn't be in `kubernetes`.   This is more kuttl specific behavior

kensipe (changes requested):

pkg/test/utils/kubernetes.go
jlpedrosa: I can move that function to the file you tell me. the thing is `kubernetes.go` is already coupled with "harness" namespace and all the utils are using the data model from API. If it would be a 1:1 mapping I would suggest yo move it to `api` package, but since I'm adding some "logic" (background false, timeout..) to make it compatbible with the existing method signatures, I am not clear what would be the "right" place. Maybe in Steps.go. 
 My natural automatic reaction would be make a pakage with all the run utils and it's own data structures, but that is clearly beyond the scope of this PR.
Let me know what is your choice.

kensipe: yeah... I need to dig in.... I was hoping not to add any more coupling there.  It's not too significant... there is a lot of refactoring I would like to do... I don't want to place that burden on you or this PR

kensipe, line 1076: I had not thought of the solution this way... but makes sense.
what doesn't make sense is the method name `toAssertCommands` it isn't `to` assert it is `from` assert commands and to standard commands.    Naming is hard but this one is misleading.   perhaps `convertAssertCommand`

pkg/test/utils/kubernetes.go
```diff
@@ -1071,6 +1071,32 @@ func RunCommand(ctx context.Context, namespace string, cmd harness.Command, cwd
...
+	commands = make([]harness.Command, 0, len(assertCommands))
+
+	for _, assertCommand := range assertCommands {
+		commands = append(commands, harness.Command{			
+			Command: assertCommand.Command,
+			Namespaced: assertCommand.Namespaced,
+			Script: assertCommand.Script,
+			SkipLogOutput: assertCommand.SkipLogOutput,
+			Timeout: timeout,
+			//This fields will always be this constans for assertions
```
kensipe, line 1086: misspelling `constans`

pkg/test/step.go
```diff
@@ -355,13 +355,24 @@ func (s *Step) CheckResourceAbsent(expected runtime.Object, namespace string) er
 	return nil
 }
 
+// CheckAssertCommands checks if the commands defined in TestAsserts are successfully running
+func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, command []harness.TestAssertCommand, timeout int) ([]error) {
+	testErrors := []error{}	
+	if _, err :=  testutils.RunAssertCommands(s.Logger,namespace, command, "", timeout ); err != nil {
```
kensipe, line 361: In order to understand this, it is necessary to dig deep.. it could be useful to add a comment that the errors is a combination of errors to execute (command doesn't exist) AND exit error codes.    Just a suggestion... otherwise looks good.

pkg/test/step.go
```diff
@@ -402,8 +413,11 @@ func (s *Step) Run(namespace string) []error {
 		return testErrors
 	}
 
-	for i := 0; i < s.GetTimeout(); i++ {
-		testErrors = s.Check(namespace)
+	timeoutF := float64(s.GetTimeout())
```
kensipe, line 416: this needed some attention.. is there a reason to change it on this PR?  could you open this separately please.

kensipe (changes requested):
/lgtm. Nice @jlpedrosa   The only thing stopping this is tests... we need to create tests which are often serve as an example of how to do things.

consider adding to:  https://github.com/kudobuilder/kuttl/tree/main/pkg/test/step_integration_test_data

pkg/test/utils/kubernetes.go
jlpedrosa: Renamed

pkg/test/utils/kubernetes.go
jlpedrosa: Corrected

pkg/test/step.go
jlpedrosa: Comment added

pkg/test/step.go
jlpedrosa: Let me explain why this is I consider this necessary in this PR. As commands for assertions most likely will be far less than instant, the previous implementations won't respect the timeout at all. Let's assume the command takes 10 seconds to execute, and the timeout is 30 seconds, this means that the loop would be running for 330 seconds (10*30 + 1*30) as each loops executes it (10 secs) and just waits 1 second.

kensipe: yeah... I've been frustrated with the "meaning" of timeout and our inconsistent use of timeout for awhile.  I imagined it needed to be address.  I just thought we would have some discussion around it.. etc.    The problem with it in this PR is it is an important component lost in the noise of another important component.  It establishes an opinion which should be debated a bit.

Things to figure out:
1. what does timeout apply to... it could be the testsuite timeout, it could be a test, or a test step.  It could be each independent element in the teststep...
2. The timeout could be separate for the apply phase and assert phase or be inclusive
3. When a timeout change is made it is common to have it made for the assert phase (with knowledge that it takes extra time for the apply to be realized in the assert).  Does that affect assert only.  Is it all asserts collectively.

I also imagine that AssertCommands, might make sense to have a timeout themselves (which is currently missing).  There is the length of time allow for a successful invocation, there is the number of times you are willing to make the invocation (defined here.. as the number of times it can be called within an assert timeout.).

Timeouts can be very important for eventually consistent tests... I believe we have a number of timeouts conflated conceptually together and we need to clean up what a timeout means at different levels of usage.

I will look past it here.  Your position makes sense.   I'm really excited with contributors add value to through PRs as you have and look to provide guidance... but also want to make sure the burden isn't high.  Thanks for the contribution!

pkg/test/step_integration_test.go
```diff
@@ -430,3 +430,100 @@ func TestStepFailure(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	errors := step.Run("irrelevant")
+	assert.Equal(t, len(errors), 1)
+}
+
+func TestAssertCommandsShouldTimeout(t *testing.T) {
+	
+	step := &Step{
+		Name:  			"commandRun",
```
kensipe, line 509: our linter must not be picking up the integration tests... but these are not properly formatted

kensipe (changes requested):
This is close... it isn't properly formatted for the integration tests... otherwise I'm good.  It would be great to get @alenkacz review as well

I will look to understand how to make sure integration builds are included in the linting

pkg/test/step_integration_test.go
kensipe: @jlpedrosa making the mod temporarily to the make from this PR will help ensure alignments on tests.  https://github.com/kudobuilder/kuttl/pull/274

kensipe:
@jlpedrosa is this still  WIP?  or ready for review?  Committers are more likely to review if WIP is complete.  thanks again!

pkg/test/step.go
```diff
@@ -402,13 +417,18 @@ func (s *Step) Run(namespace string) []error {
...
+	start := time.Now()
+
+	for elapsed := 0.0; elapsed < timeoutF; elapsed = time.Since(start).Seconds() {
```
alenkacz, line 423: I need to take a closer look at this, I'll circle back to this on Monday. I would expect we have a context set with that timeout and then we pass context around also including all the methods.

Maybe this is valid implementation as well and we can do the context one as follow-up though.

alenkacz (changes requested):
Generally looks good. Please bare with me with regards to the timeout stuff :)

pkg/test/step.go
```diff
@@ -355,14 +356,28 @@ func (s *Step) CheckResourceAbsent(expected runtime.Object, namespace string) er
 	return nil
 }
 
+// CheckAssertCommands checks if the commands defined in TestAsserts have been run successfully
+// the errors returned can be a a failure of executing the command or the failure of the command executed.
+func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, command []harness.TestAssertCommand, timeout int) []error {
```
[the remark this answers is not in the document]
jlpedrosa: I agree, but it was inconsistent with all the other `ChecksXXXX`  that were public ie: `CheckResource` `CheckResourceAbsent`.  Let me know you decision, and I'll adapt the code.

pkg/test/step.go
jlpedrosa: I think that would be better, and that is the direction I was going (that is why I added ctx to many methods). But also i was trying not to do too much code changes to try to adhere to @kensipe guidelines. My reasoning was: if it's a good practice and it has minimal  or no impact, the do it. (That is why adding the context fell in).

pkg/test/step.go
jlpedrosa: I think I get your point,  but I would say the modification of the code, through maybe opinionated, has a very small scope and it's not taking a big step in any direction. In other words, what ever discussion and design choice made later does not get impacted by applying this PR or not, the diff of this code and the previous one is minimal. Of course that is my opinion from outside,  what ever you guys decide I'll get it implemented.

pkg/test/utils/kubernetes.go
```diff
@@ -1083,7 +1109,7 @@ func RunCommands(logger Logger, namespace string, commands []harness.Command, wo
 
 	for i, cmd := range commands {
 
-		bg, err := RunCommand(context.Background(), namespace, cmd, workdir, logger, logger, logger, timeout)
+		bg, err := RunCommand(ctx, namespace, cmd, workdir, logger, logger, logger, timeout)
```
alenkacz, line 1112: just wondering what motivated this change? Was it that originally you planned to pass in context from the top but in the end did not end up doing it and considered it out of scope of this PR?

pkg/test/step.go
```diff
@@ -355,14 +356,28 @@ func (s *Step) CheckResourceAbsent(expected runtime.Object, namespace string) er
...
+	testErrors := []error{}
+	if _, err := testutils.RunAssertCommands(ctx, s.Logger, namespace, command, "", timeout); err != nil {
+		testErrors = append(testErrors, err)
+	}
+	return testErrors
+}
+
 // Check checks if the resources defined in Asserts and Errors are in the correct state.
```
alenkacz, line 369: let's also not that we run commands now

pkg/test/step.go
alenkacz, line 361: command -> commands (it's an array)

alenkacz (approved):
Thanks, in the end I have only couple of nits. I want to take a closer look at the timeouts after this merges anyway. With timeout on step level and also on the command level and context being passed in but not containing any timeouts, it's a bit not straightforward but that's not your fault :)

Thanks for the great contribution 👏

pkg/test/step_integration_test.go
jlpedrosa: @kensipe  if you could merge #274 I could rebase mine and run the test. If I cherry pick that commit here we would have conflicts no?

alenkacz: merged :)

pkg/test/utils/kubernetes.go
jlpedrosa: Yes, I was preparing to do bigger changes on the timeouts (as you suggested) but the other conversations in the ticket seemed that it would be too big of a change, so I left what made sense on it's own and was a good golang practice (pass the context).
NExt week I could give another hand for the global timeouts if you would like.

pkg/test/step_integration_test.go
jlpedrosa: rebased, and tests fmtd.

pkg/test/utils/kubernetes.go
alenkacz: Sounds great, thank you :)

pkg/test/step.go
jlpedrosa: Fixed.

jlpedrosa: fixed.

pkg/test/step.go
jlpedrosa: I am not sure what you ment here. I've updated the comment as what I read that the comment was not explaining that we also run the commands here.

pkg/test/step.go
kensipe: one of the timeouts... should be passed thru with context... completely agree with @alenkacz .   not holding this PR to that... but is the right way to manage 1 or 2 of the timeouts.  a topic which should tackle next if possible.

kensipe (approved):
/lgtm

Good stuff!  Thanks for making kuttl better!!

PR #281, opened 2021-03-09
Add tar and gzip to the image.

chhsia0:

This PR adds the `tar` and `gzip` utilities to the kuttl image, so users could run installation scripts for other tools (e.g., https://helm.sh/docs/intro/install/#from-script) needed in their tests.

Follow-up question: is it necessary to have `vim` in the image?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN make cli
 # release image with kubectl + kuttl
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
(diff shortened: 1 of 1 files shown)


kensipe:
vim should be everywhere :).   good question!    thanks for the PR and raising the questions.   I will follow up on this shortly.

kensipe (approved):
lgtm

We will remove vim as well in another PR... thanks for the awareness!

alenkacz (approved):
Thanks

PR #289, opened 2021-04-23
Flush command log always

### LINKED ISSUE #287: TestAssert command does not show command output, but TestStep does
Be able to see command output even when using TestAssert

**Why is this needed**:
Good for debugging tests

alenkacz:
Currently, we flush only if ANY collector is specified. I think it would make more sense to flush always. If we take a look at how is the step logger used, there are logs like "starting to process step" or the actual output of the assert command. I don't think this should be linked to specifying collector if I don't want any other logs collected.

Fixes #287


kensipe (approved):
lgtm

PR #266, opened 2021-01-25
Support multiple Kubernetes clusters by supporting kubeconfig in test steps

jbarrick-mesosphere:
Adds a `kubeconfig` setting to the TestStep API allowing an alternative
kubeconfig path to be used for asserts, errors, commands and applies in a test step.

Implements KEP-0008 (#261)

Signed-off-by: jbarrick@mesosphere.com <jbarrick@mesosphere.com>

PR #291, opened 2021-05-12
Support multiple Kubernetes clusters by supporting kubeconfig in test steps - Follow up for #266

takirala:

Follow up for #266

Implements KEP-0008 (#261)

### DIFF
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -99,6 +99,9 @@ type TestStep struct {
 
 	// Allowed environment labels
 	// Disallowed environment labels
+
+	// Kubeconfig to use when applying and asserting for this step.
+	Kubeconfig string `json:"kubeconfig,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -209,8 +209,8 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 		clients[testStep.Kubeconfig] = cl
 	}
 
-	for _, client := range clients {
-		if err := t.CreateNamespace(client, ns); err != nil {
+	for _, c := range clients {
+		if err := t.CreateNamespace(c, ns); err != nil {
 			tc.Failure = report.NewFailure(err.Error(), nil)
 			test.Fatal(err)
 		}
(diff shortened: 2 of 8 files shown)


kensipe (commented):
@takirala awesome sauce.   there is a minor nit that would be nice to resolve or get feedback on.   I need to take a more exhaustive review on... which I will have time tomorrow to do.

the concern is...   the kubeconfig file in the cwd is created for a test... it is tied to 1 cluster.  it is used in part by commands in tests...   if we have multiple clusters in the test.. what is that tied to and does it change.  and at what scope is it changeable?   Do we allow a change at the case? or a step?  a suite?    If allowed to change at the case level which is what it looks like here...  what is responsible for resetting it?  is it clean at the end of a case?  or does the start of a new case reset it?

I'll go back and reread the KEP to see if that was captured.. but also need to understand how that is captured in code here.

The extended issue is... it is conceivable that cluster 1 is the main under test cluster which is marked as don't teardown... but we might need to teardown cluster 2.  I'm not sure we want both clusters to be tied to the teardown flag.  It is responsible to expect that 1 cluster we don't have the rights to create namespaces but in the other cluster we don't etc.   We don't have to solve all those issues in one PR.. just need to think it through and not paint us in a corner.

takirala:
Thanks for the review @kensipe 

The kubeconfig file is limited by scope to the TestStep ([as described in the kep](https://github.com/kudobuilder/kuttl/pull/261/files#diff-187c159c0eefcfae086ce91df3cd8222f07c4be5cf7d56fae7da42762fb82b5dR48). This is both intentional and desirable - for example, to write a test that has multiple steps where in each step is interacting with a unique cluster (currently I use this for testing stuff in multiple federated clusters and we have to keep switching contexts as the tests progress).  

> If allowed to change at the case level which is what it looks like here... what is responsible for resetting it? is it clean at the end of a case? or does the start of a new case reset it?

The kubeconfig is automatically reset (to global default) at the end of the test case. (This is expected and current behavior)

> .. but we might need to teardown cluster 2

This was not captured in the KEP and also come to think of it, there may be easily more than 2 clusters (as is the case in my current use case). The global default is the only cluster the teardown flag is responsible for and all other clusters lifecycle is left to the user (Kuttl  neither creates the additional cluster(s) nor deletes them). 

> We don't have to solve all those issues in one PR.. just need to think it through and not paint us in a corner.

Agreed! I was sheperding the work done by original authors here but as far as I can see i don't see ourself putting in a corner, but happy to amend as you feel fit.

kensipe (approved):
thanks for all the details and walking thru it.  Makes sense.

/lgtm

kensipe:
good stuff @takirala thanks for pushing this across the finish line.

PR #299, opened 2021-06-22
Try fetching v1 events if v1beta1 events are not available.

porridge:
**What this PR does / why we need it**:

Makes logging events work on GKE.
Ugly (code duplication), but simple.

Fixes #298

PR #300, opened 2021-06-22
Try fetching v1  or corev1 events if v1beta1 events are not available. #299

### LINKED ISSUE #298: Fetching events does not work on GKE stable
**What happened**:

Events were not fetched when running against a GKE cluster:
```
Failed to collect events for mytest in ns mynamespace: no matches for kind "Event" in version "events.k8s.io/v1beta1"
```

Events should be printed, like they are when running against built-in docker-for-mac k8s cluster on my laptop.

1. Start a GKE cluster
2. Run any failing test against it.

**Environment**:
- Kubernetes version (use `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.6-gke.1000", GitCommit:"3ae0998c5052f420a17cb96bacf860ec5d6822a3", GitTreeState:"clean", BuildDate:"2021-04-29T09:17:16Z", GoVersion:"go1.15.10b5", Compiler:"gc", Platform:"linux/amd64"}` I also tried it against `1.18.17-gke.1901` with the same result.
- KUTTL version (use `kubectl kuttl version`): `KUTTL Version: version.Info{GitVersion:"0.10.0", GitCommit:"fc8c0f2", BuildDate:"2021-05-19T13:16:03Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}`
- Cloud provider or hardware configuration: GKE

porridge:
What this PR does / why we need it:

Makes logging events work on GKE.
Ugly (code duplication), but simple.

This supersedes #299 (that one does not seem to do the trick on GKE 1.18).

Fixes #298

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -145,6 +145,19 @@ func (o byFirstTimestampV1) Less(i, j int) bool {
 	return o[i].ObjectMeta.CreationTimestamp.Before(&o[j].ObjectMeta.CreationTimestamp)
 }
 
+// byFirstTimestampCoreV1 sorts a slice of corev1 by first timestamp, using their involvedObject's name as a tie breaker.
+type byFirstTimestampCoreV1 []corev1.Event
+
+func (o byFirstTimestampCoreV1) Len() int      { return len(o) }
+func (o byFirstTimestampCoreV1) Swap(i, j int) { o[i], o[j] = o[j], o[i] }
+
+func (o byFirstTimestampCoreV1) Less(i, j int) bool {
+	if o[i].ObjectMeta.CreationTimestamp.Equal(&o[j].ObjectMeta.CreationTimestamp) {
+		return o[i].Name < o[j].Name
+	}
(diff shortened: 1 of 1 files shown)


kensipe:
@porridge interest issue... thanks for tracking it down.   If I read the comments in the issue correctly... it seems we could just use the `v1` events.   Is there a reason to not just switch to v1 and avoid v1beta1 issues?

porridge:
> it seems we could just use the `v1` events. Is there a reason to not just switch to v1 and avoid v1beta1 issues?

@kensipe it's just compatibility concerns:
1. I'm not sure which k8s versions support the `v1` events
2. even if `v1` events are supported by a given API server, it is conceivable that events that were written using the old API versions are not readable via `v1`

That's why I went for a fall-back rather than just switch to `v1`.

kensipe:
make sense to me!  thanks for all the work!  lets merge it!

kensipe (approved):
lgtm

PR #303, opened 2021-06-24
fix: handle kubeconfig path resolution in steps properly

### LINKED ISSUE #302: Using a custom kubeconfig inside TestStep does not work always
**What happened**:

I created a teststep that looks like:

```
apiVersion: kuttl.dev/v1beta1
kind: TestStep
kubeconfig: ../../../kubeconfig-${CLUSTER_NAME}
...
```

The kubeconfig should be resolved to the path that i intended. It gets expanded first (environ substitution and then resolving the relative path) and then prepended again. It resolved to.

```/home/user/myproject/home/user/myproject/kubeconfig-cluster``` while i was expecting it to resolve to
```/home/user/myproject/kubeconfig-cluster```

Create a simple test step with a kubeconfig location as above.

reproducible irrespective of environment

**Environment**:
- Kubernetes version (use `kubectl version`):
- KUTTL version (use `kubectl kuttl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### PR DESCRIPTION
**Pull request by takirala**

When specifying absolute paths for kubeconfig in test spec, the `kubeconfig` field is [cleaned first (relative path is expanded and converted to absolute path)](https://github.com/kudobuilder/kuttl/blob/fc8c0f2b2f338dc39de1ff346077eee175ed4172/pkg/test/step.go#L521) and then that path is prepended again (irrespective of it being an absolute or relative path). 

Fixes #302

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -1016,7 +1016,11 @@ func RunCommand(ctx context.Context, namespace string, cmd harness.Command, cwd 
 	kuttlENV["NAMESPACE"] = namespace
 	kuttlENV["KUBECONFIG"] = fmt.Sprintf("%s/kubeconfig", actualDir)
 	if kubeconfigOverride != "" {
-		kuttlENV["KUBECONFIG"] = filepath.Join(actualDir, kubeconfigOverride)
+		if filepath.IsAbs(kubeconfigOverride) {
+			kuttlENV["KUBECONFIG"] = kubeconfigOverride
+		} else {
+			kuttlENV["KUBECONFIG"] = filepath.Join(actualDir, kubeconfigOverride)
+		}
 	}
 	kuttlENV["PATH"] = fmt.Sprintf("%s/bin/:%s", actualDir, os.Getenv("PATH"))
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by gracedo — APPROVED** [review]

File: pkg/test/utils/kubernetes.go [hunk 1]
```diff
@@ -1016,7 +1016,11 @@ func RunCommand(ctx context.Context, namespace string, cmd harness.Command, cwd
 	kuttlENV["NAMESPACE"] = namespace
 	kuttlENV["KUBECONFIG"] = fmt.Sprintf("%s/kubeconfig", actualDir)
 	if kubeconfigOverride != "" {
-		kuttlENV["KUBECONFIG"] = filepath.Join(actualDir, kubeconfigOverride)
+		if filepath.IsAbs(kubeconfigOverride) {
+			kuttlENV["KUBECONFIG"] = kubeconfigOverride
+		} else {
+			kuttlENV["KUBECONFIG"] = filepath.Join(actualDir, kubeconfigOverride)
```
kensipe, line 1022: this looks off in Go code... it is uncommon to see nested if statements.  It would be better IMO to have a function that takes actualDir and override and returns the proper kubeconfig... or it could just take the override and acquire the actualdir in the function body.   This would simplify this assignment and put the logic in a testable function.  it would also make it more idiomatic go, as the function would just return a value when it is known, resulting code being more left indented.   It's a nit and non-blocker.  it would be better but marginallyl

kensipe (approved):
thanks @takirala !   IMO there is a more idiomatic way which would be more testable... but not blocking.

alenkacz (approved):
LGTM, thanks

takirala:
Thanks a lot @kensipe !

PR #320, opened 2021-09-10
Correct --config help string.

porridge:

For example running `kubectl-kuttl-0.11.0 test tests/upgrade` in a directory with `kuttl-test.yaml` does not return in an error. Instead it loads the config and then overrides the `testDirs` with the command line argument. This works regardless of whether an explicit `--config kuttl-test.yaml` is specified or not.
```
2021/09/09 09:42:02 kutt-test config testdirs is overridden with args: [ tests/upgrade ]
```
This is a common convention and quite useful.


kensipe (approved):
woot... lgtm!   thank you!

PR #326, opened 2021-10-01
Create a folder when it doesn't exist and is needed for the report

### LINKED ISSUE #325: The test report will not be generated if the folder to save the report doesn't exist
**What happened**:
When the `--report=xml` option is provided, if the folder where the report will be created doesn't exist, `kuttl` fails. This is the error output:
```sh
=== CONT  kuttl
    harness.go:399: run tests finished
    harness.go:508: cleaning up
    harness.go:563: removing temp folder: ""
    harness.go:508: cleaning up
    harness.go:563: removing temp folder: ""
    harness.go:589: fatal error writing report: open tests/_build/artifacts/kuttl-test.xml: no such file or directory
--- FAIL: kuttl (62.16s)
        --- PASS: kuttl/harness/awesometest (1.22s)
```

The folder should be created

Run a test, provide the `--report=xml` parameter and don't create the folder where the reports will be generated.

iblancasa:
Signed-off-by: Israel Blancas <iblancas@redhat.com>

Fixes #325

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"time"
 )
@@ -216,6 +217,15 @@ func latestEnd(start time.Time, testcases []*Testcase) time.Time {
 // Report prints a report for TestSuites to the directory.  ftype == json | xml
 func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	ts.Close()
+
+	// Create the folder to save the report if it doesn't exist
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		dirCreationErr := os.MkdirAll(dir, 0755)
+		if dirCreationErr != nil {
+			return dirCreationErr
+		}
+	}
+
 	// if a report is requested it is always created
 	switch ftype {
 	case XML:
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -219,11 +219,15 @@ func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	ts.Close()
 
 	// Create the folder to save the report if it doesn't exist
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
-		dirCreationErr := os.MkdirAll(dir, 0755)
-		if dirCreationErr != nil {
-			return dirCreationErr
+	_, err := os.Stat(dir)
+
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(dir, 0755)
+		if err != nil {
+			return err
 		}
+	} else if err != nil {
+		return err
 	}
 
 	// if a report is requested it is always created


iblancasa:
One of [the contributing steps](https://github.com/kudobuilder/kuttl/blob/main/CONTRIBUTING.md#contributing-steps) is:
> If your proposed change is accepted, and you haven't already done so, sign a Contributor License Agreement (see details above).

Where can I sign it? Thanks! :relaxed:

kaiwalyajoshi:
@iblancasa :wave: 
Many thanks for your contributions.

The CLA you need to sign is found [here](https://github.com/kubernetes/community/blob/master/CLA.md)
I'll update the docs accordingly to add a link to the CLA.

iblancasa:
Hi @kaiwalyajoshi, it seems I fixed the problem :) Thanks. Do I need to sign anything else? If not, can we modify the "Contributing Steps" from `CONTRIBUTING.md` to remove this?:
> If your proposed change is accepted, and you haven't already done so, sign a Contributor License Agreement (see details above).

pkg/report/report.go
```diff
@@ -216,6 +217,15 @@ func latestEnd(start time.Time, testcases []*Testcase) time.Time {
 // Report prints a report for TestSuites to the directory.  ftype == json | xml
 func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	ts.Close()
+
+	// Create the folder to save the report if it doesn't exist
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		dirCreationErr := os.MkdirAll(dir, 0755)
```
kensipe, line 223: it would be great if we could shorten the name `dirCreationErr` to much shorter.   `creationErr` is better but we could also reuse the `err` here... if is throw away at this point anyway.

pkg/report/report.go
```diff
@@ -216,6 +217,15 @@ func latestEnd(start time.Time, testcases []*Testcase) time.Time {
 // Report prints a report for TestSuites to the directory.  ftype == json | xml
 func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	ts.Close()
+
+	// Create the folder to save the report if it doesn't exist
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
kensipe, line 222: I'm concerned we may be missing / ignoring certain errs.  This is only handling the error "not exist" but there are other error that are possible.  This is an area of frustration in Go.. but it is common.

See:  https://github.com/kudobuilder/kuttl/blob/main/pkg/http/client.go#L75-L82

kensipe (changes requested):
At a min, could you shorten the err var name.  and consider if other errors are important to return (which will drive if it is reasonable to "reuse" the `err` var.

I'll be quick on re-review.  Thanks for the contribution!

pkg/report/report.go
iblancasa: Sure!

pkg/report/report.go
iblancasa: I see. How about returning the error? Because things like not having permission or something can not be solved from the program, though. If the file already exists we can overwrite it but I don't know if it is a good idea to do this silently.

iblancasa:
@kensipe thanks for your review!

kensipe (approved):
/lgtm

I'm somewhat anti- elseif in go.   what do you think of:
```go
	_, err := os.Stat(dir)
	switch  {
	case os.IsNotExist(err):
		err = os.MkdirAll(dir, 0755)
		if err != nil {
			return err
		}
	case err != nil:
		return nil
	}
```

I'm good either way.  Expecting that a day will give thinking / response time.  Will merge in either state tomorrow.

kensipe:
and thanks for the contribution!   for awareness... Looking to do a release end of this week with go 1.17 bump and this change.

iblancasa:
@kensipe awesome! Thanks

Issue #203, opened 2020-08-20
$NAMESPACE substitution does not happen with arbitrary resources


joejulian:
**What happened**:
I tried to create a ClusterRoleBinding for the ServiceAccount created using the kuttl provided namespace:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: kuttlcrb
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: kuttlaccount
  namespace: $NAMESPACE
```
The resource was created exactly as printed, with `$NAMESPACE` for the namespace.

**What you expected to happen**:
I mistakenly read https://kuttl.dev/docs/testing/steps.html to say that in _any_ of the preceding topics on that page, `$NAMESPACE` would be replaced with the namespace.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KUTTL version (use `kubectl kuttl version`): KUTTL Version: version.Info{GitVersion:"0.6.0", GitCommit:"7b0360a", BuildDate:"2020-08-10T18:22:27Z", GoVersion:"go1.14.6", Compiler:"gc", Platform:"linux/amd64"}

kensipe:
kuttl was designed to work within a namespace... originally it created a test namespace for a test then deleted that namespace.  It has evolved to allow for the running within a specified namespace in which case it doesn't own that namespace.
Put another way... kuttl was built to work on namespaced resources and when used in this way, it is best to leave the namespace off the resource... kuttl will do the right thing.

We haven't taken cluster-wide resources (as in this example) into consideration yet (from a design standpoint).   
It is also worth pointing out that kuttl doesn't yet parse / augment the manifest runtime objects... we don't read or process the content (which means we won't "see" the $NAMESPACE variable).  I have been thinking about adding labels... which means augmenting the runtime objects.   This is going to take some thinking and some infrastructure buildout.

I re-read the https://kuttl.dev/docs/testing/steps.html#running-commands The references to $NAMESPACE, $PATH, etc is under the running commands section and is specific to that.  Perhaps we need to further clarification around that... it would certainly help to see an example on that page. 

[...]

PR #329, opened 2021-10-25
manifest envsubst

so-jelly:

This PR replaces variable references in manifests with environment variables.   It also adds the generated pet name to the environment so that it can be templated as well.

Fixes # https://github.com/kudobuilder/kuttl/issues/203


kensipe (changes requested):
I'll take a deeper look later today... however I would move out of `test/utils`.   It was there prior to my getting involved but have always disliked it.   We don't use "utils" in Go generally.  the package name provides meaning and context to the use of a func or struct.   Could we please consider moving it to a more meaningful package?

We are not going to introduce a need copyright into our code base tho... so translate.go needs to change to a deps or be written.

kensipe:
Thanks for the contribution...   We need a sign-off on the DCO to move forward

gberche-orange:
thanks @so-jelly for this contrib! It would very much helm me with asserting cluster-wide resources dynamically instanciated by a kuttl and referencing the associated $NAMESPACE

Any chance for you to proceed with the DCO signature so that this feature can be merged,, such as running `git rebase --signoff HEAD~2` ?

haarchri:
@so-jelly  can you rebase and add DCO - happy to see this change comes in ;)

so-jelly:
I refactored to remove that licensed code. Some edge cases i considered is variables in GVK and keys.   I have thought through this a bit and I think it is sound.   I can add some test cases as well.

so-jelly:
I created test cases for some additional field entries.  I am not sure what is up with Github review requests. tagging @kensipe in case the review request isn't sufficient.

jaypipes (approved):
We need this basically yesterday :) Any chance we can see this merged?

pkg/test/case.go
```diff
@@ -386,6 +386,10 @@ func (t *Case) determineNamespace() *namespace {
 		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
 		ns.AutoCreated = true
 	}
+
+	if _, ok := os.LookupEnv("NAMESPACE"); ok { // user can supply this environment variable
```
jaypipes, line 390: should this be 
```suggestion
	if _, ok := os.LookupEnv("NAMESPACE"); !ok { // user can supply this environment variable
```
otherwise the user's existing NAMESPACE environment variable will be overridden...

jaypipes (changes requested):

pkg/test/case.go
so-jelly: lgtm

kensipe:
@jaypipes I can provide details elsewhere... I'm just seeing this.   looks urgent and apologies for being out of pocket... for way too long.  I will review today/night and respond and/or merge

dooman87:
Wonder if this is something that could be added. 

Our use case is we need a random string in the resource manifest so external resources are not clashing when running.

porridge:
@dooman87 there is https://github.com/kudobuilder/kuttl/pull/387 that I have higher hopes for... but not enough time to work on it further at the moment.

As for random strings, have you investigated `generateName` ([docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values))?

dooman87:
Thanks @porridge, #387 looks promising! 

I looked into `generateName` but could you use it for any field? I have something like:

```yaml
apiVersion: ourcrd/v1beta1
kind: CustomObject
metadata:
  name: test-object
spec:
  path: /path/to/s3/need-random-value-here/
```

Because every test is running it its own namespace having the same name is fine, but we need different path for every test run.

I workarounded it by generating manifest and using `envsubst` but it does look hacky :)

porridge:
No @dooman87 , it's specific to metadata.name I think.

Believe me, I know the pain of `envsubst` 😅

PR #334, opened 2022-02-14
mod: Upgrade to go 1.17

haoxins:


kensipe:
appreciate the contribution!   there are 2 different `require` sections in the go.mod... which we don't want.   That said... I'm needing this for other reasons.   I think I will merge to give some credit... but update to resolve.  Thanks for the bump!

kensipe:
interesting... perhaps there is a change I wasn't familiar with...  the 2 separate require blocks for direct and indirect.   cool... disregard my previous comments... this looks good.  looks like I have a new thing to read up on.  thanks again!

kensipe (approved):
lgtm

kensipe:
for anyone else for which this is new:  https://go.dev/doc/go1.17#go-command

> If a module specifies go 1.17 or higher, the module graph includes only the immediate dependencies of other go 1.17 modules, not their full transitive dependencies. [...]

> [...] If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

PR #345, opened 2022-03-10
KEP-0008: Multi-cluster - Proposals for CLI Changes

Miles-Garnsey:
**What this PR does / why we need it:**

KEP-0008 is currently light on detail. To get it ready for implementation this PR fleshes out proposed changes to the CLI to support running multi-cluster tests using kuttl.


kensipe (commented):
this seems like a good start and expectations.  It does seem that we may want more config capabilities in the future... but I like this proposal!   the `--kind-***`  flags are for automated startup of kind... and choosing out for multi-cluster makes a lot of sense!

I wonder if there is work regarding `--namespace`  as well... do we expect all clusters to have the defined namespace?  the namespace was added based on cluster demand for cluster for which namespace creation is not allowed.

Miles-Garnsey:
> I wonder if there is work regarding --namespace as well... do we expect all clusters to have the defined namespace? the namespace was added based on cluster demand for cluster for which namespace creation is not allowed.

To clarify, you mean the `--namespace` parameter passed to kubectl? 

I think that we should assume that either (1) the namespace exists on all clusters or (2) the namespace can be created on all clusters (if `namespace` is unset in `TestSuite` and is to be auto-created).

We could allow setting different namespaces per cluster (via `TestStep` configurations or via a comma separated list in the CLI options) but it probably isn't necessary for a MVP?

PR #346, opened 2022-03-16
Bump to Go 1.18

kensipe:

Fixes #


kensipe:
looks like golangci-lint is having an issue..  we may need to delay 1.18 move

kensipe:
https://github.com/golangci/golangci-lint/issues/2374

kensipe:
more details here: https://github.com/golangci/golangci-lint/issues/2649

PR #348, opened 2022-03-24
Enable receiving a rest config on TestSuite

### LINKED ISSUE #347: Enable passing rest config to KUTTL
We have been running our integration tests using KUTTL, and we are currently trying to run them in another environment, in which  there are some requirements in order to connect to the cluster. Before connecting, we need to handle tunneling and an SSH helper, which is done by setting a custom dial function on the rest config. With this, we can create a clientset to connect to the cluster.
However, KUTTL can't receive that clientset nor the rest config to connect. While taking a look at how KUTTL works, I saw that it also uses a rest config in pkg/test/harness.go, but it's not an exported variable. The Config() function has three options to create the config, but if a config is already set, it would be reused.
I tried running a test while making changes to a local KUTTL in which the test suite can receive a rest config, and Config() checks for its existence. This test ended up being able to connect to the cluster.

alifelan:
Signed-off-by: Ali Felan <alifelan@google.com>

This adds an extra option to TestSuite, which allows KUTTL to run with a rest config, opening possibilities for customization on Go.

This is still missing testing and documentation, but wanted to share it to get feedback. My first try was making the config an exported name in harness, but that required a lot of changes in names (config to Config and Config() to GetConfig()). Adding it to TestSuite was easier, but it's the only non primitive type there.

Fixes #347

### DIFF
--- a/pkg/test/harness.go
+++ b/pkg/test/harness.go
@@ -238,6 +238,9 @@ func (h *Harness) Config() (*rest.Config, error) {
 
 	var err error
 	switch {
+	case h.TestSuite.Config != nil:
+		h.T.Log("running tests with passed rest config.")
+		h.config = h.TestSuite.Config
 	case h.TestSuite.StartControlPlane:
 		h.T.Log("running tests with a mocked control plane (kube-apiserver and etcd).")
 		h.config, err = h.RunTestEnv()
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -3,6 +3,7 @@ package v1beta1
 import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -67,6 +68,8 @@ type TestSuite struct {
 	Namespace string `json:"namespace"`
 	// Suppress is used to suppress logs
 	Suppress []string `json:"suppress"`
+
+	Config *rest.Config `json:"config,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/test/harness_integration_test.go
+++ b/pkg/test/harness_integration_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	harness "github.com/kudobuilder/kuttl/pkg/apis/testharness/v1beta1"
+	testutils "github.com/kudobuilder/kuttl/pkg/test/utils"
 )
 
 func TestHarnessRunIntegration(t *testing.T) {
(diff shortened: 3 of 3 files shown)


kensipe:
interesting @alifelan and thank you!   Apologies for missing the submitted issue which provided some of the context.  let's continue a conversation on the issue and return here.   It's a fairly small PR.. just need to understand the consequences of it.  Does this PR resolve the env issue for you?

alifelan:
Hey @kensipe! No problem, hope it makes more sense with the context in the issue, the PR is actually really small. But yeah, by doing this we can have KUTTL connect to the cluster using the rest config.

pkg/test/harness.go
```diff
@@ -236,6 +236,11 @@ func (h *Harness) Config() (*rest.Config, error) {
 		return h.config, nil
 	}
 
+	if h.TestSuite.Config != nil {
```
kensipe, line 239: while this looks similar to the quick exit above...  it isn't the same IMO.   I would rather the `switch` below be the ways in which the `h.config` is configured... more importantly we should `h.T.Log` this.

kensipe (commented):
@alifelan ok... enough thinking... more doing (on my part).  I can't think of an issue here... it is an optional element.     As noted below,  I have a preference to have config selection fall into the same logic.. additionally this will not immediately return but will create a `kubeconfig` file which is used for some of the commands if included in the kuttl tests.

It would be great to have a test for this need feature as well.   
This change + a test for the new configuration and we can land this.

Is that something you are up for?  or should I merge and create a PR to cover it?

alifelan:
Sounds great! I updated it, now the check is in the switch, and also added a test. I'm looking for feedback on that, since I tried creating a test environment using `testutils.StartTestEnvironment`, and from there passing the config to harness. The problem with that is that the mocked controlplane is skipped on the service account check, otherwise raises an error. To achieve that, I set `StartControlPlane` to true and changed the order of the switch. I don't feel that's the perfect solution, since setting `StartControlPlane` in there is a bit of a lie, but I didn't want to add a config just for this specific use case. If you have any other solution, please let me know!

kensipe (approved):
/lgtm

Thanks so much!

kensipe:
the CI doesn't seem to run for non-committers... I don't remember the details around this... all standard CI activities tested good.  Ready to merge.   Adding a todo to enable CI for all PRs

kensipe:
For awareness.. this broken main on merge...  I should have caught it on review... but I blame the lack of CI builds for non-committers as the issue.   The `make generate` is called along with `verify-generate.sh` in the CI.  It rarely changes and is rarely called by developers.   We need the safety net of the CI.   I will submit a PR fix for the break.

kensipe:
@alifelan any thoughts on re-attempting this PR?   I hated reverting it.. but need main to be buildable.

alifelan:
Hey @kensipe! I've been taking a look, though I don't have a solution yet. Seems like `TLSClientConfig` is a promoted field in `rest.Config`, and it's `DeepCopy` method is the one being used (`rest.Config` has a `CopyConfig` method, but that won't be used by the deepcopy-gen). The only idea I have is moving the field from TestSuite to Harness, but I don't think that's the way to go from a design point of view. I'll think about it, and share my ideas with you!

PR #364, opened 2022-04-21
Test Report Dir Bug Fix

kensipe:
A newly added feature was to create the directory for test reports if it didn't exist.  The default if not specified location of the test report was "", which equated to the current directory.  The check on "" existence was false which lead to the `mkdir ""` which is obvious going to fail.   The solution is to verify that the artifactsDir isn't "" prior to that logic.

Thanks to the community Erik and @iblancasa  for helping discover and track down so quickly.  We will be getting a patch release out today. 

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -218,16 +218,11 @@ func latestEnd(start time.Time, testcases []*Testcase) time.Time {
 func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	ts.Close()
 
-	// Create the folder to save the report if it doesn't exist
-	_, err := os.Stat(dir)
-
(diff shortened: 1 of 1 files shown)


iblancasa (approved):

pkg/report/report.go
```diff
@@ -241,6 +234,20 @@ func (ts *Testsuites) Report(dir, name string, ftype Type) error {
 	}
 }
 
+func ensureDir(dir string) error {
+	if dir != "" {
```
erikgb, line 238:
```suggestion
	if dir == "" {
```

erikgb (approved):
LGTM

Issue #371, opened 2022-05-17
Show content of unexpected resources


porridge:
**What would you like to be added**:

Just like `kuttl` shows a diff between expected and actual resource for `assert.yaml` files, it should show the content of the actual resource when an `errors.yaml` says that the resource should not exist (i.e. when the expected resource specifies just `apiVersion`, `kind` and `metadata.name`).

**Why is this needed**:

Currently all that is printed is:

```
    case.go:363: resource matched of kind: authorization.openshift.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: authorization.openshift.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: authorization.openshift.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
    case.go:363: resource matched of kind: /v1, Kind=ConfigMap
```

Which is not terribly useful - in order to investigate the cause for unexpected existence of a resource I'd like to see its `metadata` (to see owner references for example) and `status` (which might show details of why it was not deleted).

FTR, [this the place where the message is emitted](https://github.com/kudobuilder/kuttl/blob/fc8c0f2b2f338dc39de1ff346077eee175ed4172/pkg/test/step.go#L360).

PR #380, opened 2022-08-22
Improve error message for test assertions

porridge:
# What this PR does / why we need it:

With a test case such as this:
```
[mowsiany@mowsiany kuttl]$ cat ns.yaml 
apiVersion: v1
kind: Namespace
status:
  phase: Active
[mowsiany@mowsiany kuttl]$ kubectl-kuttl errors --timeout 1 ns.yaml 
```

The result is:

## Before this change:

```
resource matched of kind: /v1, Kind=Namespace
Error: error asserts not valid
```

which is not terribly useful for debugging..

## After this change:

```
resource /v1, Kind=Namespace default (and 65 other resources) matched error assertion
Error: error asserts not valid
```

Getting information about the name of the matched resource was instrumental for solving the issue described in https://github.com/porridge/config-race-repro

See descriptions of individual commits for details.

Fixes: #371


iblancasa (approved):

kensipe (approved):
LGTM.  thanks @porridge!

kensipe:
apologies @porridge.  my delay in getting to PRs had me merge an earlier PR which had collisions on the dependency in go.mod.   I will handle the conflict if you don't have time.   Ready to merge after all is resolved

Issue #350, opened 2022-04-01
Allow Kuttl to build a test matrix based on multiple test dimensions


soenkeliebau:
**What would you like to be added**:

_Problem_
Currently Kuttl does not allow creating something like a test matrix which would allow testing any combination of a number of parameters.
For example testing an operator that rolls out a product would require copy and pasting test cases for every version that is under testing. This becomes even worse when the test is dependent on multiple dimensions, for example product version and java version.

Consider as an example the following dimensions:

Product Version:
- 0.1.0
- 0.2.0
- 0.2.1
- 0.3.0

Java Version:
- 8
- 11
- 16

This would require copy and pasting the same test case 12 times and adjusting the values manually in the test files, which is error prone, laborious and hard to maintain.

_Proposal_

We would like to be able to define test dimensions for Kuttl tests which can then be used to parameterize tests and effectively create a test matrix at runtime.

We have implemented external tooling around Kuttl at the moment to provide this functionality based on a test dimensions file that has the following structure:

````yaml
---
dimensions:
  - name: spark
    values:
      - 3.2.1
      - 3.3.0
  - name: hadoop
    values:
      - 3.2.0
      - 3.2.1
  - name: stackable
    values:
      - 0.4.0
tests:
  - name: spark-pi-private-s3
    dimensions:
      - spark
      - hadoop
      - stackable
  - name: spark-pi-public-s3
    dimensions:
      - spark
      - hadoop
      - stackable
````

This allows defining an arbitrary number of test dimensions along with values, as well as test cases that use these dimensions and which dimensions they actually use.
Our tooling then creates the needed list of test cases and runs a templating engine over the input kuttl files to generate the actual test files into those test case directories.

For the above example we'd generate the following list of test cases:

- spark-pi-private-s3
  - spark-3.2.1_hadoop-3.2.0_stackable-0.4.0
  - spark-3.3.0_hadoop-3.2.0_stackable-0.4.0
  - spark-3.2.1_hadoop-3.2.1_stackable-0.4.0
  - spark-3.3.0_hadoop-3.2.1_stackable-0.4.0
- spark-pi-public-s3
  - spark-3.2.1_hadoop-3.2.0_stackable-0.4.0
  - spark-3.3.0_hadoop-3.2.0_stackable-0.4.0
  - spark-3.2.1_hadoop-3.2.1_stackable-0.4.0
  - spark-3.3.0_hadoop-3.2.1_stackable-0.4.0
 
**Why is this needed**:
[...]

Issue #288, opened 2021-04-22
Asserts should be able to reference the namespace that kuttl has created


waciumawanjohi:
**What would you like to be added**:
An ability to refer to the namespace in which tests are running.

I suspect this would look like some templating. e.g.:
```yaml
status:
  thingRef:
    namespace: $(NAMESPACE)
```
Alternatively, I could imagine this being implemented as a comment with special meaning to kuttl:

```yaml
status:
  thingRef:
    namespace: # //NAMESPACE//

```

**Why is this needed**:
I have resourceA and resourceB:
`00-setup.yaml`
```yaml
apiVersion: waciuma.com/v1alpha1
kind: thing
metadata:
  name: resource-a

---

apiVersion: waciuma.com/v1alpha1
kind: watcher
metadata:
  name: resource-b
```
I want to assert that resourceB's status contains an ObjectReference to resourceA. Currently, I can write the following:
`00-assert.yaml`
```yaml
apiVersion: waciuma.com/v1alpha1
kind: watcher
metadata:
  name: resource-b
status:
  thingRef:
    name: resource-a
```

I want to further assert that status.thingRef.namespace is correct.

I can not do this currently, unless I specify a namespace for the tests to run in.

I don't want to hardcode into the test suite what namespaces will be used in the tests. This will make tests brittle and more fragile moving from one environment to another. I cannot guarantee that another developer will have created a special namespace in the cluster that they are using for tests.

Related to #203

faiq:
This comes up as an issue when creating service accounts in namespaces and associating role bindings with them. 

The only solution I've come up with is running the steps in a dedicated namespace as you mentioned and deleting that namespace in a different test step.

Issue #262, opened 2021-01-11
Add ability to match/assert fields on regex values


chrisguest75:
**What would you like to be added**:
I would like the ability to partially match fields of the resources in the asserter using something akin to a regex.   

This is predominantly for integration/production test scenarios where you are not in complete control of the set values. But understand the range they should operate within.  

**Why is this needed**:
The ability to do this will allow us to validate api-resource fields such as; resource limits,  image tags, custom annotations and other fields that do not require discrete values to be valid for a test pass.

erikgb:
I would also like to see this implemented - even if my use case is a bit simpler. As described on [Slack](https://kubernetes.slack.com/archives/CG3HTFCMV/p1616958778013400), my wish is to be able to **assert that a field is present**, but with no asserts on the field value.

We built an operator on top of [cert-manager](https://cert-manager.io/), and as part of our e2e testing, we would like to assert that the certificate Secret contains the expected fields. But we can ofc not include any asserts on the fields - as the fields in question are transient: private keys, certificate chains, keystores etc.

I am not sure how this feature could be implemented - since a field may have _any_ value. One idea I have, is to follow the [ytt](https://carvel.dev/ytt/) approach, and express these kinds of asserts with "magic comments" as field values - in the test step asserts. This might look like this:

```yaml
---
kind: Secret
apiVersion: v1
metadata:
  name: my-certificate
data:
  tls.crt: #@ regex('^(?!\s*$).+')
type: kubernetes.io/tls
```

@alenkacz WDYT?

DanArlowski:
Second this, 
configmaps and stuff that are autogenerated may be great to assessing using regex,

right now you cant asses a configmap value if some of it is autogenerated

haarchri:
any updates here ?

porridge:
This is possible with [CEL expression support in TestAssert](https://github.com/kudobuilder/kuttl/blob/main/docs/testing/reference.md#testassert).
FTR kuttl now has basic [templating support](https://github.com/kudobuilder/kuttl/pull/642).

Issue #224, opened 2020-10-12
Allow assert of a key inside a resource without defining the actual value


ANeumann82:
**What would you like to be added**:
A way to assert that a key in a resource is set, but without actually defining the value of that key - normally, because it will contain a semi-random value.

**Why is this needed**:
For example to check if cert-manager has inserted the `caBundle` into a resource. The caBundle value will be semi-random, so we can't put the value inside the `00-assert.yaml`, but we still want to make sure that they key is set:

```
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  name: cert-manager-webhook
webhooks:
  - clientConfig:
[...]

PR #387, opened 2022-09-13
[WiP] Support template expansion of YAML files

porridge:
# Why do we need it

I encountered a few cases where I needed `kuttl` to match a field to a value that depends on the execution environment. Examples:
- product version string in a `status` field in order to verify that an upgrade succeeded. The version string is produced based on git commit SHA dynamically by the build tooling.
- cluster-scoped resources whose name includes the name of the test namespace

I ended up creating a `TestAssert` containing a script which:
1. does poor man's template expansion using `envsubst`
2. invokes a (nested) `kuttl assert` on the resulting file
3. deletes the temporary assert file

As you can imagine, the result is not pretty or straightforward to understand at first sight.

Other issues that I found which might (or might not) be helped by this change: https://github.com/kudobuilder/kuttl/issues/70 https://github.com/kudobuilder/kuttl/issues/160 https://github.com/kudobuilder/kuttl/issues/350 https://github.com/kudobuilder/kuttl/issues/288 https://github.com/kudobuilder/kuttl/issues/262 https://github.com/kudobuilder/kuttl/issues/224 https://github.com/kudobuilder/kuttl/issues/222 https://github.com/kudobuilder/kuttl/issues/203

# What this PR does

During a recent team hackathon I took a stab at finding out how difficult it would be to do the template expansion directly within `kuttl`. Turns out it was not too difficult, and the result works. But please keep in mind this is a *hack*, not meant to be merged as is. But might be a useful inspiration on the necessary work.

For backwards compatibility I introduced a notion of `*.gotmpl.yaml` files, and only these are expanded. The dictionary of expansion keys (`TemplatingContext`) also needs some more thought. For now it consists of the namespace and environment.

### DIFF
--- a/pkg/file/files.go
+++ b/pkg/file/files.go
@@ -13,11 +13,11 @@ import (
 )
 
 // from a list of paths, returns an array of runtime objects
-func ToObjects(paths []string) ([]client.Object, error) {
+func ToObjects(paths []string, templatingContext testutils.TemplatingContext) ([]client.Object, error) {
 	apply := []client.Object{}
 
 	for _, path := range paths {
-		objs, err := testutils.LoadYAMLFromFile(path)
+		objs, err := testutils.LoadYAMLFromFile(path, templatingContext)
 		if err != nil {
 			return nil, fmt.Errorf("file %q load yaml error: %w", path, err)
 		}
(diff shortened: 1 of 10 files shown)


porridge:
Obsoleted by https://github.com/kudobuilder/kuttl/pull/642

Issue #76, opened 2020-05-07
Allow partial comparison of slices


alembiewski:
**What would you like to be added**:

It would be great to check only the subset of a slice in a test step assertion. Currently, it is required to specify all the array elements, otherwise, a test fails with `slice length mismatch` error.

**Why is this needed**:

Here are some real-life pain points:
- if istio injection is enabled, a sidecar is running alongside with the main container in a pod, and it needs to be reflected in test assertions;
- testing mutating webhook, which updates environment variables or volume/volume mounts for some container in a pod, and a test case checks the applied patch;
- pod's status conditions (say, the goal of a test step to check the only one, final condition);
- and so on

kensipe:
@alembiewski could you provide an example of what is required today and what you would like to see (yaml examples will help visualize).  This seems valuable.

kensipe:
we need enough info to create a kep

alembiewski:
Here is the example: let's say we have a custom resource `TrainingJob`, which describes some machine-learning training job, and it has a `status.conditions[]` field (slice). The job goes through several conditions before it reaches the final state. In my test case, I want to check the job is succeeded. Currently, to work-around the `slice length mismatch` error, my assert file looks like this:
```
kind: TrainingJob
metadata:
  name: ml-job
status:
  conditions:
  - reason: JobCreated      # I have to provide all the intermediate conditions here
  - reason: JobRunning      # to make kuttl happy
  - reason: JobSucceeded   <-- this is what I actually want to test
    status: "True"
    type: Succeeded
```
This is how I would like my assert to be:
```
kind: MLJob
metadata:
  name: ml-job
status:
  conditions:
  - reason: JobSucceeded
    status: "True"
    type: Succeeded
```
So the problem is that I have to specify all the conditions (but this could be env variables, volumes - any property with array type).

kensipe:
This is related to #75 
Looking for better array support in asserts.   Looking for:
1. Disregard order to array
2. Partial array support (only assert what is expressed)

porridge:
Please note that the workaround mentioned in https://github.com/kudobuilder/kuttl/issues/76#issuecomment-632112138 is fragile: we've seen cases where the elements appear in a different order, and this causes the test to fail.

nic-hima:
[...]

PR #388, opened 2022-09-13
[WiP] Assertions based on strategic merge patch (partial list comparisons)

porridge:
# Why we need  this

IMHO one of the most pressing issue in `kuttl` is the problem with assertions on lists (arrays, slices, whatever). To match a list:
- all elements must be listed
- they must be exactly in the same order

For example with `status.conditions`, depending on the operator which writes the list, this might be impossible to achieve (e.g. if ordering is random or some conditions are present or absent inconsistently.

See #76 for details.

# What this PR does

During a recent team hackathon I took a stab at finding out how easy it would be to use the strategic merge patch machinery in order to relax the requirements mentioned above. We basically treat the assertion as a strategic merge patch, apply it to the object at hand, and the assertion passes if and only if the resulting object does not differ from the original.

Credit for the idea goes to @misberner

To be completely explicit, this is just a prototype, but can serve as an inspiration to someone who wanted to take this further.

### DIFF
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/kube-openapi/pkg/util/proto"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


porridge:
I think more testing and cleaning up the code would be necessary.

Since this has plenty of potential to trip people up in some corner cases I didn't think about, we should probably consider some more gentle rollout strategy than just wholesale replacing the old comparison mechanism with the new one. Perhaps a toggle in a `TestAssert`? :thinking:

chipzoller:
We (Kyverno) would be very interested in having this. The Policy Report format we use outputs an array/object within the CR and that ordering is not guaranteed. We will want to be able to test that a single object that looks like _X_ definition occurs somewhere in that list.

anderssonw:
We would also love for this to work. We have loads of partial list comparisons in our tests, which for some reason only sometimes fails, probably due to some race condition we are not interested in testing for the tests where we get fails.

haarchri:
@porridge can we help with something ? we very interested out of crossplane project in this feature ;)

porridge:
> @porridge can we help with something ? 

I think [these comments](https://github.com/kudobuilder/kuttl/pull/388#issuecomment-1285901412) still apply.

PR #395, opened 2022-10-26
Allow user to set reportName

### LINKED ISSUE #284: ReportName is hard coded to "kuttl-test" due to which we are not able to provide customize name for the report
https://github.com/kudobuilder/kuttl/blob/6de9356d102ff0db4234598762c9b38dad44fe84/pkg/kuttlctl/cmd/test.go#L114

**reportName** parameter in test suite yaml is not taking customized name as it is hardcoded in the test.go, kindly make that as a configurable parameter in test suite yaml. As we are running multiple suites one by one so all the reports generated by a previous suites are getting overwritten by the next suite.

**Sample Test Suite Yaml :**

```
apiVersion: kuttl.dev/v1beta1
kind: TestSuite
startKIND: true
kindContainers:
- your/image:latest
testDirs:
- tests/e2e/
reportName: abc-test
reportFormat: JSON
```

Thanks
Vikas

### LINKED ISSUE #394: Allow `reportName` to be set with a flag

Currently, KUTTL 0.13.0 doesn't seem to honor `reportName` in the test suite configuration file, cf. https://github.com/kudobuilder/kuttl/issues/284.

On top of fixing that issue, it would be nice to be able to set the reportName through the CLI flags, to allow the configuration file to be overriden.

**Why is this needed**:

When running a series of KUTTL tests, the hard-coded file name will result in overwriting; this can be gotten around by parameterizing the `--artifacts-dir` flag with the name of the test series.

For example, if running tests `1..5`, this would result in `1/kuttl-test.json`, `2/kuttl-test.json`, etc.

That's OK, but it means that the test parameter is only available in the file path; for human readability, it would be nice to have the name of the test in the report name.

benjaminjb:

* change default report name to `kuttl-report` -- I can nix this part if you don't want it, but it seems to me that `kuttl-test.json`/`kuttl-test.xml` is too close to `kuttl-test.yaml`, the default config name, and liable to lead to some confusion.
* override default report name according to config file's `reportName` field
* override default or config file report name according to command line option --report-name

Signed-off-by: Ben Blattberg <ben.blattberg@crunchydata.com>

Fixes #394 
Fixes #284

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -58,6 +58,7 @@ func newTestCmd() *cobra.Command {
 	mockControllerFile := ""
 	timeout := 30
 	reportFormat := ""
+	reportName := "kuttl-report"
 	namespace := ""
 	suppress := []string{}
 
@@ -111,7 +112,6 @@ For more detailed documentation, visit: https://kuttl.dev`,
 			}
 
 			// Override configuration file options with any command line flags if they are set.
-			options.ReportName = "kuttl-test"
 			if isSet(flags, "crd-dir") {
 				options.CRDDir = crdDir
 			}
@@ -169,6 +169,10 @@ For more detailed documentation, visit: https://kuttl.dev`,
 			if isSet(flags, "report") {
 				var ftype = report.Type(strings.ToLower(reportFormat))
 				options.ReportFormat = reportType(ftype)
+			}
+
+			if isSet(flags, "report-name") {
+				options.ReportName = reportName
 			}
 
 			if isSet(flags, "artifacts-dir") {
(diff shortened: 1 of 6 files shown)


pkg/kuttlctl/cmd/test.go
```diff
@@ -58,6 +58,7 @@ func newTestCmd() *cobra.Command {
 	mockControllerFile := ""
 	timeout := 30
 	reportFormat := ""
+	reportName := "kuttl-report"
```
benjaminjb, line 61: These vars are set here, but it seems like the main value of setting the default value here is as a reminder of what the default is, since they get overwritten by the config file, yeah?

pkg/test/harness.go
```diff
@@ -602,11 +602,20 @@ func (h *Harness) Report() {
...
 
+// GetReportName returns the configured ReportName.
+func (h *Harness) GetReportName() string {
+	reportName := "kuttl-report"
+	if h.TestSuite.ReportName != "" {
+		reportName = h.TestSuite.ReportName
+	}
+	return reportName
+}
+
```
benjaminjb, line 618: Since this pattern was already established by the Timeout field, I followed that method of setting the default where it would be used.

kensipe:
thanks @benjaminjb !   I will take a closer review tomorrow and we may need to coordinate with our operator sdk scorecard folks...    lets hold until we get their feedback... but that shouldn't take long.

jmrodri:
@benjaminjb @kensipe this change will definitely break [`operator-sdk`](https://github.com/operator-framework/operator-sdk/) but I think it is a good change. `kuttl-report` is a more appropriate name IMO. And I am definitely in favor of adding a flag to allow folks to change it. So `operator-sdk` can easily just change to using the flag and we'll always be okay.

We'll probably want to flag this as a 'breaking' change when `kuttl` is released.

jmrodri (approved):
/lgtm

pkg/kuttlctl/cmd/test.go
kensipe: after some refactoring, I understand better your question.  IMO we should refactor the init to the be the order defined and not "check" the value throughout code.

Issue #397, opened 2022-10-31
Automatic cleanup/deletion of cluster-scoped CR


chipzoller:
Not sure if this is a bug, feature request, or just operator error. In running `kuttl` I'm finding it doesn't remove a cluster-scoped CustomResource at the conclusion of a test suite. `skipDelete` is set at the default. I also see that a Namespaced resource, as part of the test suite, is also not cleaned up. I'm new to kuttl and I might have missed something super basic. Appreciate any pointers.

iblancasa:
I think your issue is related to #253

chipzoller:
That issue is two years old. Is the workaround to manually perform a delete via a Command?

iblancasa:
AFAIK, yes

chipzoller:
@eddycharly does your PR #401 completely solve for this or are there still outstanding actions to close this out/

Issue #253, opened 2020-12-07
Not all created objects are deleted when using a specific namespace


cscetbon:
**What happened**:
When test is done an object is still there 
```
kgp  -n n1
NAME   READY   STATUS    RESTARTS   AGE
pod1   1/1     Running   0          86s
```

**What you expected to happen**:
I'd expect kuttl to delete all objects created during a test case, in my case pod1 and pod2, and not only pod2.

**How to reproduce it (as minimally and precisely as possible)**:
have the following structure 
multi-resources
    ├── 00-assert.yaml
    ├── 00-pod1.yaml
    ├── 01-assert.yaml
    └── 01-pod2.yaml

with the following files https://gist.github.com/cscetbon/3064597e8cb51af2d3efdd6b81a80b4c

**Anything else we need to know?**:
Here are the logs from kuttl
```
kuttl test
=== RUN   kuttl
    harness.go:447: starting setup
    harness.go:248: running tests using configured kubeconfig.
    harness.go:275: Successful connection to cluster at: https://34.77.7.231
    harness.go:343: running tests
    harness.go:74: going to run test suite with timeout of 30 seconds for each step
    harness.go:355: testsuite: . has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/multi-resources
=== PAUSE kuttl/harness/multi-resources
=== CONT  kuttl/harness/multi-resources
    logger.go:42: 23:20:21 | multi-resources | Ignoring .DS_Store as it does not match file name regexp: ^(\d+)-([^.]+)(.yaml)?$
    logger.go:42: 23:20:21 | multi-resources | Creating namespace: kuttl-test-worthy-mastiff
    logger.go:42: 23:20:21 | multi-resources/0-pod1 | starting test step 0-pod1
    logger.go:42: 23:20:23 | multi-resources/0-pod1 | Pod:n1/pod1 created
    logger.go:42: 23:20:25 | multi-resources/0-pod1 | test step completed 0-pod1
    logger.go:42: 23:20:25 | multi-resources/1-pod2 | starting test step 1-pod2
    logger.go:42: 23:20:27 | multi-resources/1-pod2 | Pod:n1/pod2 created
    logger.go:42: 23:20:29 | multi-resources/1-pod2 | test step completed 1-pod2
[...]

PR #401, opened 2022-11-08
fix: step cleanup not capturing current step

eddycharly:

This PR fixes step cleanup not capturing the current step.

Without that, cleanup is only invoked with the last step pointer (called multiple times, one per step).

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -347,11 +347,11 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 		tc.Assertions += len(testStep.Errors)
 
 		if !t.SkipDelete {
-			defer func() {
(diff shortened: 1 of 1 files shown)


kensipe (approved):
great find and contribution!  thank you!

I would normally push for a test to confirm the resolution of a bug like this... however this looks obvious and will work separate for a test.

chipzoller:
Does this PR close issues #397 and #253? Or would PR #402 address any of those?

erikgb:
> Does this PR close issues #397 and #253? Or would PR #402 address any of those?

I bet it fixes #253, not sure about #397.

eddycharly:
> not sure about https://github.com/kudobuilder/kuttl/issues/397

From what I saw in the code base it should be correctly handled.

PR #402, opened 2022-11-08
refactor: use t.Cleanup instead of defer

eddycharly:

This PR makes use `t.Cleanup` instead of `defer`.

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -317,23 +317,21 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 	}
 
 	for _, c := range clients {
+		c := c
 		if err := t.CreateNamespace(c, ns); err != nil {
 			tc.Failure = report.NewFailure(err.Error(), nil)
 			test.Fatal(err)
-		}
-	}
-
-	if !t.SkipDelete {
-		defer func() {
-			for _, c := range clients {
+		} else if !t.SkipDelete {
+			test.Cleanup(func() {
 				if err := t.DeleteNamespace(c, ns); err != nil {
 					test.Error(err)
 				}
-			}
-		}()
+			})
+		}
 	}
 
 	for _, testStep := range t.Steps {
+		testStep := testStep
 		testStep.Client = t.Client
 		if testStep.Kubeconfig != "" {
 			testStep.Client = newClient(testStep.Kubeconfig)
@@ -347,11 +345,11 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 		tc.Assertions += len(testStep.Errors)
 
 		if !t.SkipDelete {
-			defer func(step *Step) {
-				if err := step.Clean(ns.Name); err != nil {
+			test.Cleanup(func() {
+				if err := testStep.Clean(ns.Name); err != nil {
 					test.Error(err)
 				}
-			}(testStep)
+			})
 		}
 
 		if errs := testStep.Run(ns.Name); len(errs) > 0 {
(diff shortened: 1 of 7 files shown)


pkg/test/case.go
```diff
@@ -317,23 +317,21 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
 	}
 
 	for _, c := range clients {
+		c := c
```
eddycharly, line 320: This is not that nice.

pkg/test/case.go
```diff
@@ -317,23 +317,21 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
...
 					test.Error(err)
 				}
-			}
-		}()
+			})
+		}
 	}
 
 	for _, testStep := range t.Steps {
+		testStep := testStep
```
eddycharly, line 334: This is not that nice.

pkg/test/case.go
```diff
@@ -347,11 +345,11 @@ func (t *Case) Run(test *testing.T, tc *report.Testcase) {
...
 
 		if !t.SkipDelete {
-			defer func(step *Step) {
-				if err := step.Clean(ns.Name); err != nil {
+			test.Cleanup(func() {
+				if err := testStep.Clean(ns.Name); err != nil {
 					test.Error(err)
 				}
-			}(testStep)
+			})
```
eddycharly, line 352: Maybe this should go in `testStep.Run` ?

pkg/test/case.go
```diff
@@ -94,6 +94,14 @@ func (t *Case) CreateNamespace(cl client.Client, ns *namespace) error {
...
 	}
 
+	if !t.SkipDelete {
+		test.Cleanup(func() {
+			if err := t.DeleteNamespace(cl, ns); err != nil {
+				test.Error(err)
+			}
+		})
+	}
+
```
eddycharly, line 104: This looks good, we now can register ns cleanup at the same place where we create it.

pkg/test/case_integration_test.go
```diff
@@ -21,14 +21,14 @@ func TestMultiClusterCase(t *testing.T) {
 		t.Error(err)
 		return
 	}
-	defer testenv.Environment.Stop()
+	t.Cleanup(testenv.Environment.Stop)
```
kensipe, line 24:
```suggestion
	t.Cleanup(func() {
		if err := testenv.Environment.Stop(); err != nil {
			t.Error(err)
		}
	})
```

This doesn't compile... param is `func() {` and `Stop` is of type `func() error {` this was likely missed due to this file being `go:build integration`

pkg/test/case_integration_test.go
```diff
@@ -21,14 +21,14 @@ func TestMultiClusterCase(t *testing.T) {
...
 
 	testenv2, err := testutils.StartTestEnvironment(testutils.APIServerDefaultArgs, false)
 	if err != nil {
 		t.Error(err)
 		return
 	}
-	defer testenv2.Environment.Stop()
+	t.Cleanup(testenv2.Environment.Stop)
```
kensipe, line 31:
```suggestion
	t.Cleanup(func() {
		if err := testenv2.Environment.Stop(); err != nil {
			t.Error(err)
		}
	})
```

kensipe (changes requested):
generally like the changes... additional changes in `case_integration_test.go` required to make it compilable and runnable.  Otherwise I think we are good to go

pkg/test/case_integration_test.go
eddycharly: Hmmm, shouldn't we at least try to compile in CI ?

kensipe (approved):
/lgtm

PR #405, opened 2022-11-09
feat: add expectFailure support to TestStep

eddycharly:

This PR adds `expectFailure` support to TestStep.
When `expectFailure` is set to `true` we expect that the manifest in the `apply` stanza will fail to apply.

NOTE: this PR contains changes from #402, the diff will be easier to read once this merges.

### DIFF
--- a/pkg/test/case_integration_test.go
+++ b/pkg/test/case_integration_test.go
@@ -22,13 +22,22 @@ func TestMultiClusterCase(t *testing.T) {
 		return
 	}
 	t.Cleanup(testenv.Environment.Stop)
+	t.Cleanup(func() {
+		if err := testenv.Environment.Stop(); err != nil {
+			t.Error(err)
+		}
+	})
 
 	testenv2, err := testutils.StartTestEnvironment(testutils.APIServerDefaultArgs, false)
 	if err != nil {
 		t.Error(err)
 		return
 	}
-	t.Cleanup(testenv2.Environment.Stop)
+	t.Cleanup(func() {
+		if err := testenv2.Environment.Stop(); err != nil {
+			t.Error(err)
+		}
+	})
 
 	podSpec := map[string]interface{}{
 		"restartPolicy": "Never",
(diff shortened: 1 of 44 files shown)


erikgb:
Having the ability to expect failure is something I would love to see. But I think we have to make some iterations on the UX. I would also like to have the ability to do assertions on error messages. For e2e-testing validating webhooks in operators. Maybe a KEP is needed?

pkg/test/case_integration_test.go
```diff
@@ -21,6 +21,7 @@ func TestMultiClusterCase(t *testing.T) {
 		t.Error(err)
 		return
 	}
+	t.Cleanup(testenv.Environment.Stop)
```
eddycharly, line 24:
```suggestion
```

iblancasa:
I agree with @erikgb. The documentation needs to be updated as well.

eddycharly:
I think I came up with a better design IMHO.
Something like this:
```yaml
apiVersion: kuttl.dev/v1beta1
kind: TestStep
apply:
- path: policy.yaml
  shouldFail: true
assert:
- policy-assert.yaml
```
Not perfect yet because it's on a per file basis and I would prefer it to be per resource though.

erikgb:
> I think I came up with a better design IMHO. Something like this:
> 
> ```yaml
> apiVersion: kuttl.dev/v1beta1
> kind: TestStep
> apply:
> - path: policy.yaml
>   shouldFail: true
> assert:
> - policy-assert.yaml
> ```
> 
> Not perfect yet because it's on a per file basis and I would prefer it to be per resource though.

I think having it on the resource level will require some very exotic YAML-syntax. Asserting failures on the file level, without asserting error messages would at least be a good start! 👍 from me!

eddycharly:
> I think having it on the resource level will require some very exotic YAML-syntax

Not necessarily, it could look something like:
```yaml
piVersion: kuttl.dev/v1beta1
kind: TestStep
apply:
- path: policy.yaml
  shouldFailFor:
    - v1/Pod/my-pod
    - ...
assert:
- policy-assert.yaml
```

> How do we define "fail"? Maybe we need to have some ability to assert the returned status code?

We are using client-go, I'm not sure it exposes a status code.

eddycharly:
Closing, old one.

PR #406, opened 2022-11-09
feat: add skip test regex support

eddycharly:

This PR allows to skip tests based on a regular expression.


erikgb:
I like the idea in this PR, but we will need some tests of this new feature and doc update. 😅

eddycharly:
> we will need some tests of this new feature and doc update. 😅

Sure, if we agree on this I will add tests and docs.
I'm not going to write tests and docs if the PR is sure to be rejected ;-)

erikgb:
@kensipe I am in desperate need of this feature. 😅 WDYT?

kensipe:
👀

kensipe:
with 2 strong community members interested in this feature... I'm inclined to merge.
thoughts towards the future:
1. do we want a fail message in the logs with a message of fail?  I'm not sure we do
2. should this be an array of regex?  or force the complexity in the regex itself?
3. do we want a cli flag to support?  I take the current desire to have file control over skipped tests which are for a period of time not good tests... not a temporary situation of skip this for this run... however the flag could also be used to negate the skip for a run which seems useful
4. we need to understand the combination of `--test` and this feature.   What if the `--test` is specified for a skipped test?   at this point it would be skipped... but is that the behavior we want?

kensipe:
I'm also not sure why we wouldn't express `t.Skipped()` to provide a go test skip vs just a message that it was skipped.

iblancasa:
I'm a bit confused (sorry). How is the feature used? I have a bunch of tests where I can check how this feature behaves :)

erikgb:
> I'm also not sure why we wouldn't express `t.Skipped()` to provide a go test skip vs just a message that it was skipped.

I agree with this.

To me, it seems like we can agree that this might be a useful feature to add to kuttl, and that it will eventually get merged after some "polishing" (docs, tests, etc.). 😄 @eddycharly, are you prepared to continue on this PR, or is Kyverno investing in your fork now instead?

Here is an example of what I have to do now to temporarily disable a kuttl test: https://github.com/statnett/image-scanner-operator/tree/main/test/e2e/scenario_disabled (move the test to a "disabled" folder).

eddycharly:
Hey folks, I added support for this in our fork, I can contribute it back upstream if you want.

I like the idea of using `t.Skipped()` !

eddycharly:
Closing, old one.

PR #415, opened 2022-11-10
feat: add full name support for test case name

### LINKED ISSUE #403: [Enhancement] Output test case full path in final report

Along with the result of the test (`PASS` | `FAIL`), print the full path to the test case if known through a TestSuite file.

**Why is this needed**:

With hundreds of possible tests being executed across a complex hierarchy of directories, knowing which test failed only by the name of the test case (i.e., the parent folder holding the kuttl test files) isn't often enough. For example, something like this is printed today:

```
        --- PASS: kuttl/harness/external-metrics (25.87s)
        --- PASS: kuttl/harness/nomutateDigest-verifyDigest-norequired (5.38s)
        --- PASS: kuttl/harness/mutateDigest-noverifyDigest-norequired (3.30s)
        --- PASS: kuttl/harness/keyless-nomutatedigest-noverifydigest-required (3.92s)
```

In reality, each of those test cases are six directories deep, and they may not all even be in the same parent directory.

This enhancement is to print, in the final report output, the full path to the test cases as known based upon the TestSuite file.

eddycharly:

This PR adds support for full folder path in the test name (instead of the shorter folder name).
This is useful when test cases are organised in deep folder hierarchy.

Fixes #403

cc @chipzoller

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,22 @@
+name: "E2E tests"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  e2e-tests:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/setup-go@v2.1.4
+      with:
+        go-version: 1.18
+    - uses: actions/checkout@v2.4.0
+    - name: "Install etcd"
+      run: ./hack/install-etcd.sh
+    - name: "Install kube-apiserver"
+      run: ./hack/install-kube-apiserver.sh
+    - name: "Run E2E tests"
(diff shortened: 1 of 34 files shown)


chipzoller:
Closes #403

kensipe:
this is interesting...   First thought that comes to mind is we support testsuites and tests that come from a tarball.   The tarballs are expanded into a system temp folder I believe.  This could make for a not so friendly full name.   It would be good to see in the PR a sample current and change output (abbrev with just the affected output).   I will look to do that now.

kensipe:
looks like expected output is in the issue... thank you!

kensipe:
focusing a conversation on the issue #403  instead of this PR

PR #337, opened 2022-02-17
Create CI infrastructure in GitHub Actions

iblancasa:
Signed-off-by: Israel Blancas <iblancas@redhat.com>

**What this PR does / why we need it**:
* Fixes #335: CI is not run in PRs
* Use GitHub Actions for the CI
* Add a job for the unit tests (missing in the CircleCI automation)

**Some notes**
* Create as a draft because I need to add a workflow to create releases

PR #423, opened 2022-11-15
chore: add gh workflow to run checks on PR

eddycharly:
This PR adds a github workflow to run checks on PRs.


iblancasa:
How about running the other tests too? I started that work in #337... but didn't have time for it.

PR #431, opened 2022-11-28
chore: use setup-envtest to bootstrap EnvTest

### LINKED ISSUE #428: GH Action Upgrade test K8S Version
We current run tests on 1.16 and we need to bump to a more modern version while also keeping in mind what is commonly in production environments

erikgb:

This replaces the shell script installation of `etcd` and `kube-apiserver` with [setup-envtest](https://github.com/kubernetes-sigs/controller-runtime/tree/0702a666bcf0d1e4b6607c4f5404a74ffc433674/tools/setup-envtest).

Also cleaning out some docs that are outdated by this change.

Upgrading to K8s version `1.25.0`, which is the latest release of kubebuilder-tools.

Fixes #428


kensipe:
I'm confused why this is passing tests... but it could be platform differences.  I am unable to run `integration-tests` nor `e2e`.   
the e2e fails with ARM issues:  unable to fetch checksum for requested version: unable fetch metadata for kubebuilder-tools-1.19.2-darwin-arm64.tar.gz -- got status "404 Not Found" from GCS
KUBEBUILDER_ASSETS="" ./hack/run-e2e-tests.sh

I will look into ARM options for around this timeframe.   This likely predates Mac M1.  
If we are unable to find a solution, we may need to hold on this and find a more modern solution (later version of k8s)

kensipe:
this is part of the issue with having the build tool manage these details.  It is great for evergreen, leading edge dev.   That said, I'm a fail of env-test (which was influx around M1 and over multiple versions of k8s).  I think most of that has been worked out.    I should add, I had issues local with integration tests as well... but for the same reason.

kensipe:
btw

```
./bin/setup-envtest list
(available)  v1.25.0  darwin/arm64
(available)  v1.24.2  darwin/arm64
(available)  v1.24.1  darwin/arm64
```
😞

erikgb:
Maybe upgrade K8s to 1.24.2 before merging this then?

iblancasa (approved):

kensipe:
@erikgb for awareness... I'm working to upgrade our CRDs to V1 then moving to the latest version of testenv.  Once this is in place, I believe we will be ready to merge this.

erikgb:
@kensipe Thanks for fixing the requirements to go beyond K8s version 1.19! Somehow I cannot run the integration tests locally now, even with K8s version 1.19.2. It just hangs, and that worked before rebasing the PR. 🤔 But CI is happy, so I suggest moving forward. Curious if running the integration tests works for you now?

kensipe:
the integration test works fine for me... reviewing again now.   but looks good

Makefile
```diff
@@ -133,10 +135,25 @@ endif
...
-integration-test:  ## Runs integration tests
-	./hack/run-integration-tests.sh
+integration-test: envtest  ## Runs integration tests
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" ./hack/run-integration-tests.sh
 
 # Run e2e tests
 .PHONY: e2e-test
-e2e-test: cli
-	./hack/run-e2e-tests.sh
+e2e-test: cli envtest
```
kensipe, line 143: the `e2e-test` is not a rerun of the integration tests... I think there was some confusion around this.   The e2e tests should NOT be dependent on the `envtest`.  e2e requires docker + kind.    Although the refactor of the circleci -> git action lost that.... which is why we continue to have a circleci task along with the github actions.   I'm not sure we need to solve all of that in this PR but we should revert this portion.   I will wait a day for a response... if none, I will merge and mod, then create a PR for the e2e to be on github actions (assuming it's possible... It will be an area I will need to research).

kensipe (commented):
Thanks @erikgb for this contribution... it is really great!   After (needed) updates to the CRDs etc... this works great.  On M1+ k8s 1.24+ is needed... making the 1.25.0 a great starting point.   I've added a comment to address but will merge with your attribution and mod tomorrow.  I wanted to make sure you had a day to respond (which given the holiday season may be somewhat short response time).

kensipe:
well... It looks like unintentionally mislead...   I'm not sure when the e2e changed.   it is using envtest but in a dockerized env.   that doesn't seem necessary and we do want a test env with kind for a real e2e.   I can see why there was confusion.   At this point we should remove the e2e and create a todo to add in a real e2e.     I have to assume that we had the real e2e on the kudo side... it seems best to disconnect from kudo and create our own e2e.   more clean up to do, but will result in a cleaner project.   Please remove the e2e target... but as previously mentioned... I'm happy to do that.

Makefile
erikgb: I've removed the `e2e-test` make target.

erikgb:
@kensipe I've removed the e2e-test make target, but unsure where to put the TODO for migrating the e2e tests to real e2e tests using kind. Please advice!

kensipe:
kind of you sir!   I will create a github issue to capture it and schedule into a milestone

kensipe (approved):
/lgtm

PR #436, opened 2022-12-01
fall back to controller-runtime defaulting for the apiserver

redradrat:

For the mock controlplane, the defaulting was copy/pasted over from controller-runtime (I assume in some earlier state). As of https://github.com/kubernetes-sigs/controller-runtime/pull/1541/files, their defaulting got revamped, and thus we can fall back to their defaulting. Only for advanced users, we can allow to override this, but should overhaul this config in the future as well, as the property used to override is deprecated.

Fixes #

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -213,9 +213,8 @@ For more detailed documentation, visit: https://kuttl.dev`,
 			var err error
 			if mockControllerFile != "" {
 				APIServerArgs, err = testutils.ReadMockControllerConfig(mockControllerFile)
-			} else {
-				APIServerArgs = testutils.APIServerDefaultArgs
 			}
+
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


kensipe:
looks like a number of PRs coming working on the same area of code... I will review this today.  thanks @redradrat !

kensipe:
I was just recently upgrading testenv libs and was debating on this... I was debating on if we should detect version and configure based on that providing tests for different versions of APIs...  It seems reasonable to move on, establish ourselves on the latest and handle any issues if they come up... it is unlikely  and the only reason I can think of to support older versions is for v1beta1 support of CRDs... but that was so long ago now.   Thanks for this PR.   I will merge this and adjust my current code accordingly.   Agree with the the overhaul on config at some point.   Thanks so much!

kensipe (approved):
I'm not sure the reason to not refactor the `StartTestEnv` function to remove the first argument... but this accomplishes the same affect.   I will follow up after this merges.  Thanks again!

redradrat:
@kensipe tbh I don't know the reason either... I wrote this so brain-fried after long hours, I even forgot sending it 😅 sorry

Issue #297, opened 2021-06-09
kubectl kuttl test leaves garbage kubeconfig file around


porridge:
**What happened**:

After `kubectl kuttl test` runs, a `kubeconfig` file appears in current directory.

**What you expected to happen**:

No garbage left behind.

**How to reproduce it (as minimally and precisely as possible)**:

```
$ cat kuttl-test.yaml 
apiVersion: kuttl.dev/v1beta1
kind: TestSuite
testDirs:
- ./empty
$ ls kubeconfig 
ls: kubeconfig: No such file or directory
$ kubectl kuttl test --namespace blah
=== RUN   kuttl
    harness.go:457: starting setup
    harness.go:248: running tests using configured kubeconfig.
    harness.go:285: Successful connection to cluster at: https://kubernetes.docker.internal:6443
    harness.go:353: running tests
    harness.go:74: going to run test suite with timeout of 30 seconds for each step
    harness.go:365: testsuite: ./empty has 0 tests
=== RUN   kuttl/harness
=== CONT  kuttl
    harness.go:399: run tests finished
    harness.go:508: cleaning up
    harness.go:563: removing temp folder: ""
--- PASS: kuttl (1.22s)
    --- PASS: kuttl/harness (0.00s)
PASS
ls -l kubeconfig
-rw-r--r--  1 marcin  staff  5484 Jun  9 12:24 kubeconfig
$
```

**Anything else we need to know?**:

👋🏻 

**Environment**:
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:15:20Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```
[...]

PR #437, opened 2022-12-02
Use the correct kubeconfig when running in-cluster

### LINKED ISSUE #393: When `kuttl` runs in a pod, `kubectl` run by it sees incorrect $KUBECONFIG
**What happened**:

When `kuttl` runs in a pod on kubernetes cluster `A`, but with the `KUBECONFIG` variable in its environment, pointing to cluster `B`:
- correctly applies manifests and checks assertions against cluster `B`,
- however any `kubectl` commands it runs from a `commands` or `collectors` inside a `TestAssert` seem have `KUBECONFIG` variable in their environments set to some weird value, as a result they find the kubeconfig for cluster `A` implicitly provided in the containing pod, and end up connecting to cluster `A` which is the wrong cluster

Example:

```
    logger.go:42: 06:11:18 | upgrade/20-upgrade-operator | running command: [kubectl logs --prefix -l app=central -n kuttl-test-desired-airedale --all-containers --tail=-1]
    logger.go:42: 06:11:18 | upgrade/20-upgrade-operator | W0901 06:11:18.221837    1761 loader.go:221] Config not found: /go/src/github.com/stackrox/stackrox/operator/kubeconfig
```

The child commands should be fed the same `KUBECONFIG` as the parent `kuttl` has.

1. Create two clusters, `A` and `B`.
2. Create a simple test suite that for example applies whatever resource (say a `ConfigMap`) and a `TestAssert` step which runs a `kubectl get` command for that resource
3. Run `kuttl` against the test suite in a pod on cluster `A` but feed it a kubeconfig for cluster `B`.

One can work this around for `commands` by using another variable, say `REAL_KUBECONFIG` on the top level, and then saying `KUBECONFIG=$REAL_KUBECONFIG` in the command/script. However no such workaround is possible for `collectors`. :disappointed:

Potentially related tickets:
- https://github.com/kudobuilder/kuttl/issues/297
- https://github.com/kudobuilder/kuttl/issues/331

**Environment**:
- Kubernetes version (use `kubectl version`): OpenShift 4.11
- KUTTL version (use `kubectl kuttl version`): `0.13.0`

porridge:

- See [the issue](https://github.com/kudobuilder/kuttl/issues/393) for an in-depth description of the problem.
- This PR is best viewed by looking at individual commits.
- The first three commits are cleanups and refactorings that should not change the behaviour.
- The last commit changes the code to no longer short-circuit in the in-cluster case, which fixes the aforementioned issue.

Fixes #393


porridge:
If someone wants to try this out, see https://github.com/porridge/kuttl/releases/tag/v0.14.0-kubeconfig

kensipe:
thanks @porridge !

porridge:
(rebased)

kensipe:
ok... this is next up for review.   expecting is to complete (and likely merge tomorrow)

kensipe:
@porridge just checking... does this work in cluster?    a quick review shows `testutils.InClusterConfig()` removed.   I believe that is need for operator sdk scorecard.   We unfortunately don't have tests to confirm that is still working.   I will need to dig in but the title of the PR indicates that expectation.

kensipe:
disregard... reviewing the context and issues makes things clear... Dziękuję!

porridge:
Yes, sorry forgot to mention: I did check that log collection works now, with `kuttl` running both in a pod and on my workstation.

pkg/test/harness.go
```diff
@@ -227,6 +227,8 @@ func (h *Harness) RunTestEnv() (*rest.Config, error) {
 
 // Config returns the current Kubernetes configuration - either from the environment
 // or from the created temporary control plane.
+// As a side effect, on first successful call this method also writes a kubernetes client config file in YAML format
+// to a file called "kubeconfig" in the current directory.
```
kensipe, line 231: 👏

pkg/test/harness.go
```diff
@@ -253,51 +255,54 @@ func (h *Harness) Config() (*rest.Config, error) {
...
-		if err != nil {
-			return nil, err
-		}
-		if inCluster {
-			return h.config, nil
-		}
 	}
 	if err != nil {
-		return h.config, err
+		return nil, err
```
kensipe, line 260: 👏

pkg/test/harness.go
```diff
@@ -253,51 +255,54 @@ func (h *Harness) Config() (*rest.Config, error) {
...
+		if err := h.waitForFunctionalCluster(); err != nil {
+			return nil, err
 		}
 		h.T.Logf("Successful connection to cluster at: %s", h.config.Host)
 	}
 
-	// The creation of the "kubeconfig" is necessary for out of cluster execution of kubectl
+	// The creation of the "kubeconfig" is necessary for out of cluster execution of kubectl,
+	// as well as in-cluster when the supplied KUBECONFIG is some *other* cluster.
 	f, err := os.Create("kubeconfig")
```
kensipe, line 280: I wonder if we should make this a test temp file / folder... however this is consistent with existing behavior deserves it's own PR if we want it.   Thanks for the Godocs!   really helpful!

pkg/test/harness.go
```diff
@@ -253,51 +255,54 @@ func (h *Harness) Config() (*rest.Config, error) {
...
-		return h.config, err
+		return nil, err
 	}
 
 	defer f.Close()
 
 	return h.config, testutils.Kubeconfig(h.config, f)
 }
 
+func (h *Harness) waitForFunctionalCluster() error {
```
kensipe, line 290: I wonder if this should be in kubernetes.go but it's just a refactor of code in the existing file... looks good!

pkg/test/harness.go
```diff
@@ -253,51 +255,54 @@ func (h *Harness) Config() (*rest.Config, error) {
 			return nil, err
 		}
 		h.config.WarningHandler = rest.NewWarningWriter(os.Stderr, rest.WarningWriterOptions{Deduplicate: true})
-		inCluster, err := testutils.InClusterConfig()
```
kensipe, line 256: I'm most concern with this removal... but it is likely based on code review the case of the issue you are looking to solve.   I vaguely remember needing this to satisfy the way operator SDK scorecard uses kuttl.

kensipe (approved):
/lgtm

left some comments... mostly celebrations to godocs and obvious fixes on err conditions.   As mentioned,  This solves an issue in one use case and I have concerns regarding the removal of inclusterconfig removal.   Additionally the change means `kubernetes.go::InClusterConfig` is no longer used and can be refactored out.   I'll grab that as a cleanup on another PR.

Thanks @porridge !   Ready to merge.

kensipe:
changed my mind :)   Removing the dead code for inclusterconfig makes sense to include on this PR.   merging upon CI build success.

pkg/test/harness.go
porridge: Yes, I think this should be in a temporary file (and created with an appropriate umask BTW), see also [this related issue](https://github.com/kudobuilder/kuttl/issues/297).

PR #438, opened 2022-12-07
core deps bump

kensipe:
Version updates for:
* docker
* k8s
* client-go
* testify
* cobra

This bump brings the k8s dependencies to v0.25.4


pkg/test/utils/kubernetes.go
```diff
@@ -224,7 +224,7 @@ func (r *RetryClient) Patch(ctx context.Context, obj client.Object, patch client
 // Get retrieves an obj for the given object key from the Kubernetes Cluster.
 // obj must be a struct pointer so that obj can be updated with the response
 // returned by the Server.
-func (r *RetryClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+func (r *RetryClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
 	return Retry(ctx, func(ctx context.Context) error {
 		return r.Client.Get(ctx, key, obj)
```
evenh, line 229: Shouldn't `opts` be passed to the `r.Client.Get` call like the CRUD functions?

kensipe: double checking... thanks for the watchful eye

PR #439, opened 2022-12-08
Remove use of APIServerArgs for testenv

kensipe:
The latest testenv removed the use of configuration via args.  Removing APIServerArgs from kuttl which removes the value of the flag `--control-plane-config`.   The flag now reports it is not used in the logs and is planned to be removed from kuttl completely after v0.16.0.

To view output try:  `go run ./cmd/kubectl-kuttl test --start-control-plane=true --control-plane-config=foo pkg/test/test_data`

Normally we would have a test to verify but this should shortly be removed.


erikgb (approved):
LGTM

PR #445, opened 2022-12-20
Remove e2e Testing on circleci

kensipe:
* Removal of e2e run scripts
* Removal of circleci (which was used for e2e)

We need to revisit on another PR the ability to have a docker, kind full e2e test and add it to the pipeline with a preference toward github actions.

Fixes #


erikgb (changes requested):
https://github.com/kudobuilder/kuttl/blob/main/test/Dockerfile#L20 refers to a file deleted in this PR. I think we should address this here.

kensipe:
takk! @erikgb  great eye... will do

PR #451, opened 2023-01-03
Bump K8s deps

evenh:
Updates Kubernetes dependencies in order to be compatible with downstream projects that use other dependencies that require newer `client-go`, `controller-runtime`, `controller-tools` etc.


kensipe (approved):
/lgtm

PR #464, opened 2023-03-11
Set reportFormat to lower before generating report

### LINKED ISSUE #449: reportFormat in TestSuite
Hello,

I have the following TestSuite file:
```yaml
apiVersion: kuttl.dev/v1beta1
kind: TestSuite
testDirs:
- ./tests/e2e
reportFormat: XML
namespace: mpms-tests-auto
```
The execution works fine but Kuttl continues to generate the report in JSON format.
Could you check that point ?

I use the version 0.14.0.

Rgds.

rishinair11:
Fixes https://github.com/kudobuilder/kuttl/issues/449

When `reportFormat` in `kuttl-test.yaml` is specified in **uppercase**, it is passed unchanged to `report.Report` func in `harness.go`.
It needs to be converted to **lowercase** so that the `switch` compares against valid report format types (`ftype`)

This has been taken care of here already. https://github.com/kudobuilder/kuttl/blob/f6d64c915c8dd9e2da354562c3d5c6fcf88aec2b/pkg/kuttlctl/cmd/test.go#L177

Also added unit tests for the `report.Report` func to verify the behaviour.


pkg/test/harness.go
```diff
@@ -605,7 +605,9 @@ func (h *Harness) Report() {
 	if len(h.TestSuite.ReportFormat) == 0 {
 		return
 	}
-	if err := h.report.Report(h.TestSuite.ArtifactsDir, h.reportName(), report.Type(h.TestSuite.ReportFormat)); err != nil {
+
+	reportType := report.Type(strings.ToLower(h.TestSuite.ReportFormat))
```
iblancasa, line 609: Instead of doing the change here, I think what should be modified is the `reportType` function from `pkg/kuttlctl/cmd/test.go` to store the report format in lower letters. That will help to not reproduce this problem in other places where `pkg/kuttlctl/cmd/test.go` is used.

iblancasa (changes requested):

pkg/test/harness.go
kensipe: +1 agree with @iblancasa 
as this has been awhile from PR to review (my fault)... if I don't see an update today... I will merge (for proper attribution) and modify consistent to iblancasa comments.

kensipe (commented):
nice work... I agree with iblancasa and looking for an update... if not, I will merge and mod.  thanks again

porridge:
Superseded by https://github.com/kudobuilder/kuttl/pull/721

PR #488, opened 2023-09-28
Update list to filter resources using label selectors

### LINKED ISSUE #480: Kuttl should be able to match resource on label if metadata.name is not provided
I would like to be able to match a resource based on labels when no name is provided. 

For example, the below resource will have an autogenerated uuid attached to the the name making it impossible to select. I want kuttl to fallback to selecting the resource based on labels. In this case, it would find a ResourceGroup with the label resourceGroupDev and then make the assertion. 

```yaml
apiVersion: azure.upbound.io/v1beta1
kind: ResourceGroup
metadata:
  labels:
    name: "resourceGroupDev"
spec:
  forProvider:
    location: westeurope
```

**Why is this needed**:
As of now, the only way to match a specific resource is by metadata.name, however many k8s resources use autogenerated names, for example, pods created by replica sets. Because of this, there is no way to match a specific pod and Kuttl will often select the wrong one which results in a failure.

jtucci:

Currently, If no name is specified in an assert resource, the CheckResource and CheckResourceAbsent will find all resources of with the same gvk and assert against one of the resources returned in the list. This PR is a proposal to enhance how resources are matched by filtering the resources by any included labels in the case where no name is specified. 

**Example Use Case / Scenario** 
We are using KUTTL for e2e testing for the crossplane compositions we are developing. The resources we are asserting against (managed resources) are created in a similar way to pods being created by a replica set, in that the metadata.name will contain a random uuid. In the case where there are several resources of the same gvk, we are unable to guarantee that the correct resource is asserted against which causes unexpected failures within the tests. The only other workaround is running script commands such as:
```bash
test "$(kubectl get virtualnetworkpeering.network.azure.upbound.io -o json | jq -r '.items[] |
      select(.metadata.labels.name=="virtual-network-peering-to-target") |
      .metadata.annotations["crossplane.io/external-name"]')" = "to-shoot--test--test-dev-us"`
```
This is very tedious (we have to do this for every field in the resource(s)). It also makes understanding any test failures difficult since no diff is shown, just that a command failed. 

This enhancement would give us the ability to instead assert against the yaml definition of the resource like any other test would because we can specify unique labels to identify the resources. For example, the following two resources with no name and the same gvk would be asserted against the correct resource with the matching labels 

```yaml
---
apiVersion: network.azure.upbound.io/v1beta1
kind: VirtualNetworkPeering
metadata:
  annotations:
    crossplane.io/external-name: to-shoot--test--test-dev-us
  labels:
    name: virtual-network-peering-to-target
spec:
  deletionPolicy: Delete
  forProvider:
    remoteVirtualNetworkId: >-
      /subscriptions/xxxxxxxxx/resourceGroups/shoot--test--test-dev-us/providers/Microsoft.Network/virtualNetworks/shoot--test--test-dev-us
    resourceGroupNameSelector:
      matchControllerRef: true
      policy:
        resolution: Required
        resolve: Always
    virtualNetworkNameSelector:
      matchControllerRef: true
      policy:
        resolution: Required
        resolve: Always
---
apiVersion: network.azure.upbound.io/v1beta1
kind: VirtualNetworkPeering
metadata:
  annotations:
    crossplane.io/composition-resource-name: virtual-network-peering-from-target
    crossplane.io/external-name: from-shoot--test--test-dev-us
  labels:
    name: virtual-network-peering-from-target
spec:
  deletionPolicy: Delete
  forProvider:
    remoteVirtualNetworkIdSelector:
      matchControllerRef: true
      policy:
        resolution: Required
        resolve: Always
    resourceGroupName: shoot--test--test-dev-us
    virtualNetworkName: shoot--test--test-dev-us
---
```
Fixes #480

### DIFF
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -278,10 +278,12 @@ func (s *Step) CheckResource(expected runtime.Object, namespace string) []error 
 		actual := unstructured.Unstructured{}
 		actual.SetGroupVersionKind(gvk)
 
-		err = cl.Get(context.TODO(), client.ObjectKey{
+		if err := cl.Get(context.TODO(), client.ObjectKey{
 			Namespace: namespace,
 			Name:      name,
-		}, &actual)
+		}, &actual); err != nil {
+			return append(testErrors, err)
+		}
 
 		actuals = append(actuals, actual)
 	} else {
@@ -289,13 +291,13 @@ func (s *Step) CheckResource(expected runtime.Object, namespace string) []error 
 		if err != nil {
 			return append(testErrors, err)
 		}
-		actuals, err = list(cl, gvk, namespace, m.GetLabels())
+		actuals, err := list(cl, gvk, namespace, m.GetLabels())
+		if err != nil {
+			return append(testErrors, err)
+		}
 		if len(actuals) == 0 {
 			testErrors = append(testErrors, fmt.Errorf("no resources matched of kind: %s", gvk.String()))
 		}
-	}
-	if err != nil {
-		return append(testErrors, err)
 	}
 
 	expectedObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(expected)
--- a/pkg/test/step.go
+++ b/pkg/test/step.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
@@ -230,13 +231,17 @@ func (s *Step) GetTimeout() int {
 	return timeout
 }
 
-func list(cl client.Client, gvk schema.GroupVersionKind, namespace string) ([]unstructured.Unstructured, error) {
+func list(cl client.Client, gvk schema.GroupVersionKind, namespace string, labelsMap map[string]string) ([]unstructured.Unstructured, error) {
 	list := unstructured.UnstructuredList{}
 	list.SetGroupVersionKind(gvk)
 
 	listOptions := []client.ListOption{}
 	if namespace != "" {
 		listOptions = append(listOptions, client.InNamespace(namespace))
+	}
+
+	if len(labelsMap) > 0 {
+		listOptions = append(listOptions, client.MatchingLabels(labelsMap))
 	}
 
 	if err := cl.List(context.TODO(), &list, listOptions...); err != nil {
@@ -280,7 +285,11 @@ func (s *Step) CheckResource(expected runtime.Object, namespace string) []error 
 
 		actuals = append(actuals, actual)
 	} else {
-		actuals, err = list(cl, gvk, namespace)
+		m, err := meta.Accessor(expected)
+		if err != nil {
+			return append(testErrors, err)
+		}
+		actuals, err = list(cl, gvk, namespace, m.GetLabels())
 		if len(actuals) == 0 {
 			testErrors = append(testErrors, fmt.Errorf("no resources matched of kind: %s", gvk.String()))
 		}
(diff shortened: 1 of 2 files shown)


pkg/test/step.go
```diff
@@ -10,7 +10,13 @@ import (
 	"testing"
 	"time"
 
+	harness "github.com/kudobuilder/kuttl/pkg/apis/testharness/v1beta1"
```
porridge, line 13: Please keep kuttl imports in a separate block.

porridge (changes requested):
This looks awesome! Would you mind adding some tests?

jtucci:
> This looks awesome! Would you mind adding some tests?

@porridge Sure thing! I just updated the test cases to allow for several 'actual' resources to be created and added a case to test the label matching.

pkg/test/step.go
```diff
@@ -280,7 +285,11 @@ func (s *Step) CheckResource(expected runtime.Object, namespace string) []error
 
 		actuals = append(actuals, actual)
 	} else {
-		actuals, err = list(cl, gvk, namespace)
+		m, err := meta.Accessor(expected)
+		if err != nil {
+			return append(testErrors, err)
+		}
+		actuals, err = list(cl, gvk, namespace, m.GetLabels())
```
porridge, line 292: Thanks!

Looks like linter caught a logic error here, though:

```
pkg/test/step.go:292:12: ineffectual assignment to err (ineffassign)
		actuals, err = list(cl, gvk, namespace, m.GetLabels())
		         ^
make: *** [Makefile:39: lint] Error 1
```

From a quick glance it seems that introducing `:=` [on line 288](https://github.com/kudobuilder/kuttl/pull/488/files#diff-f83daba5f9026600aa38b2b1e5a0c585535dc3355457e4762c50bdb0f6b0febdR288) caused the new `err` to shadow the one on the top level, resulting in the assignment on line 292 to be effectively lost?

Please give it some careful thought. This function is a subtle and critical part of the code, we should avoid introducing a regression here.

jtucci: good catch, sorry about that.

pkg/test/step.go
```diff
@@ -273,21 +278,27 @@ func (s *Step) CheckResource(expected runtime.Object, namespace string) []error
...
 			Name:      name,
-		}, &actual)
+		}, &actual); err != nil {
+			return append(testErrors, err)
+		}
 
 		actuals = append(actuals, actual)
 	} else {
-		actuals, err = list(cl, gvk, namespace)
+		m, err := meta.Accessor(expected)
```
porridge, line 290: I'm a little afraid that `Accessor()` might fail in case `expected` is not of a type [expected by it](https://github.com/kubernetes/apimachinery/blob/v0.28.2/pkg/api/meta/meta.go#L96) (`metav1.Object` or `metav1.ObjectMetaAccessor`) but was previously good enough for `CheckResource` (`runtime.Object`). The different apimachinery types always make my head spin. Did you perhaps look in detail at uses of this method and could confirm with some confidence that we'll be OK?

porridge (changes requested):
The error returning fix LGTM overall, but I have one worry, see inline.
Also, please take a look at the `TestCheckResourceIntegration/match_object_by_labels,_field_mismatch` that started failing.

pkg/test/step.go
jtucci: I think thats a valid concern. I would imagine that it could cause an issue in the case its not of type metav1.Object or metav1.ObjectMetaAccessor. Wouldn't this issue already exist though since `Namespaced()` is being called within `CheckResources()`, which is then calling `Accessor()` with `expected` as its argument? 

ttps://github.com/kudobuilder/kuttl/blob/main/pkg/test/utils/kubernetes.go#L338-L343

porridge: Great point, this is exactly the kind of argument I was looking for.

Sounds like it might make sense to change the `CheckResource()` signature to be more strict, then, but this is definitely not in scope of this PR.

porridge:
> [...] please take a look at the `TestCheckResourceIntegration/match_object_by_labels,_field_mismatch` that started failing.

Any ideas about this one, @jtucci ?

jtucci:
> > [...] please take a look at the `TestCheckResourceIntegration/match_object_by_labels,_field_mismatch` that started failing.
> 
> Any ideas about this one, @jtucci ?

I've been out of town but am going to take a look at it this week :)

porridge (approved):
:heart:

PR #490, opened 2023-10-09
Copy documentation over from `kuttl.dev` repo to `kuttl`

porridge:
**What this PR does**:

A an almost verbatim copy of the docs in the kuttl.dev repository `master` branch, **plus**:
- the changes from
https://github.com/kudobuilder/kuttl.dev/commit/e1979f486c5ea4acffac8ba12adc9ef97a246623
which document a feature already merged in kuttl `main`.
- update to reflect the change introduced in https://github.com/kudobuilder/kuttl/pull/488

The first commit is an exact copy of the above files. Additional commits are for minor styling changes needed to adapt to GitHub markdown flavor. This is the basic set of changes necessary. I hope additional improvements to restructure the docs and bring them up to date with the actual code will come later.

**Why we need it**:

The primary motivation is fixing the docs vs binary versioning conundrum: it was never clear to me whether the documentation on the kutt.dev website should correspond to the `kuttl` code  in `main` or to the last released version (there are advantages and disadvantages to both approaches).

This way we can store docs close to the same version of code they correspond to - the docs on `main` should always match the code on `main`. At the same time, the docs directory at a release-versioned URL will naturally correspond to the given release.

Arguably, this commit should also modify the PR template checklist to mention that docs should be updated together with source.

**Legacy docs**

As for the docs still on kuttl.dev, I think we should replace them with a redirect to the `docs` directory in this repo to prevent confusion, as soon as we revive the website deployment pipeline. The landing page on kuttl.dev can of course stay mostly as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ KUTTL is designed for testing operators, however it can declaratively test any k
 
 ## Getting Started
 
-Please refer to the [getting started guide](https://kuttl.dev/docs/) documentation.
+Please refer to the [getting started guide](docs/README.md) documentation.
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Learn more on how to engage with the KUDO community on the [community page](http
 
 ## Contributions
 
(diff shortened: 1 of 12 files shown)


kensipe (approved):
lgtm

PR #494, opened 2023-10-30
Use .Script to describe command if .Command is empty.

### LINKED ISSUE #493: Command.Script is ignored when creating error messages
**What happened**:

My test using a command with `script:` failed and `kuttl` produced this unhelpful message:

```
 command "" exceeded 72 sec timeout, context deadline exceeded
```

The error message should mention the content of `script:` somehow if `command:` was not provided.

Use a step such as:

```
apiVersion: kuttl.dev/v1beta1
kind: TestStep
commands:
- script: make fail
```

A workaround might be to use `command:` instead, but this is tricky if one wants to use shell syntax.

porridge:

Introduce a `String()` method that summarizes a command paying attention to `.Script` too, and use it in appropriate places.

While at it, also provide a command summary when it exits non-zero. Before this change, the error was:
```
    logger.go:42: 12:47:52 | basic-central/62-unset-expose-monitoring | running command: [some command and args]
    logger.go:42: 12:47:52 | basic-central/62-unset-expose-monitoring | <command output, if any>
    logger.go:42: 12:47:53 | basic-central/62-unset-expose-monitoring | test step failed 62-unset-expose-monitoring
    logger.go:42: 12:47:53 | basic-central/62-unset-expose-monitoring | collecting log output for [type==pod,label: app=central]
    logger.go:42: 12:47:53 | basic-central/62-unset-expose-monitoring | < potentially pages and pages of collector output >
    [...]
    logger.go:42: 12:47:53 | basic-central/62-unset-expose-monitoring | < potentially pages and pages of collector output >
    case.go:366: failed in step 62-unset-expose-monitoring
    case.go:368: exit status 1
```

After this change, the last line will be a more helpful:
```
    case.go:368: command "some command and args" failed, exit status 1
```

Fixes #493

### DIFF
--- a/pkg/apis/testharness/v1beta1/command.go
+++ b/pkg/apis/testharness/v1beta1/command.go
@@ -1,0 +1,51 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import "strings"
+
+func (c *Command) String() string {
+	if c.Command == "" && c.Script == "" {
+		return "(invalid command with neither Command nor Script set)"
+	}
+	if c.Command != "" && c.Script != "" {
+		return "(invalid command with both Command and Script set)"
+	}
+	if c.Command != "" {
+		return c.Command
+	}
+	return summarize(c.Script)
+}
+
+// summarize returns a short representation of a multi-line command.
+func summarize(script string) string {
+	var lines []string
+	for i, line := range strings.Split(script, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		if i == 0 && strings.HasPrefix(line, "set -") {
+			continue
+		}
+		lines = append(lines, line)
+	}
+	joined := strings.Join(lines, "\\n ")
+	if len(joined) > 70 {
+		return joined[:67] + "..."
+	}
+	return joined
+}
--- a/pkg/apis/testharness/v1beta1/command.go
+++ b/pkg/apis/testharness/v1beta1/command.go
@@ -17,6 +17,9 @@ package v1beta1
 
 import "strings"
 
+// String returns a human-readable representation of a Command.
+// In particular, when the .Script field is set, we try to omit comments
+// as well as `set -...` commands, and elide long content.
 func (c *Command) String() string {
 	if c.Command == "" && c.Script == "" {
 		return "(invalid command with neither Command nor Script set)"
@@ -30,7 +33,8 @@ func (c *Command) String() string {
 	return summarize(c.Script)
 }
 
-// summarize returns a short representation of a multi-line command.
+// summarize returns a short representation of a multi-line shell script.
+// It tries to remove comments and `set -...` commands.
 func summarize(script string) string {
 	var lines []string
 	for i, line := range strings.Split(script, "\n") {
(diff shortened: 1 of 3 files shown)


mclasmeier (approved):

pkg/apis/testharness/v1beta1/command.go
```diff
@@ -0,0 +1,51 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import "strings"
+
+func (c *Command) String() string {
```
kensipe, line 20: This is an implementation of the [stringer](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) interface and curious if it's implementation is consist or confusing to stringer expectations.   Since it is just returning a string, perhaps it is... however the summarized output makes it seem like it is different.

pkg/apis/testharness/v1beta1/command.go
```diff
@@ -0,0 +1,51 @@
...
+	if c.Command == "" && c.Script == "" {
+		return "(invalid command with neither Command nor Script set)"
```
kensipe, line 22: This seems like an error message but is part of the string.  Are we trying to avoid err messages and error handling on the calling side?   Are we going to run into case of uses where the caller is going to need to scan for these error messages?

pkg/apis/testharness/v1beta1/command.go
```diff
@@ -0,0 +1,51 @@
...
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		if i == 0 && strings.HasPrefix(line, "set -") {
+			continue
+		}
+		lines = append(lines, line)
+	}
+	joined := strings.Join(lines, "\\n ")
+	if len(joined) > 70 {
```
kensipe, line 47: `70` and `67` seem like magic numbers... it would be best to have constants or a constant that expresses a string limit.  I would recommend a string limit constant and a (constant-3) or `join[:SIZE_LIMIT-len("...") + "..."`

pkg/apis/testharness/v1beta1/command.go
```diff
@@ -0,0 +1,51 @@
...
+	if c.Command != "" && c.Script != "" {
+		return "(invalid command with both Command and Script set)"
+	}
+	if c.Command != "" {
+		return c.Command
+	}
+	return summarize(c.Script)
+}
+
+// summarize returns a short representation of a multi-line command.
```
kensipe, line 33: this method doesn't just shorten a multi-line command... in fact it is somewhat js aware and removes comments and set conditions.   It seems designed specific to scripts and not commands.

pkg/apis/testharness/v1beta1/command.go
```diff
@@ -0,0 +1,51 @@
...
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		if i == 0 && strings.HasPrefix(line, "set -") {
+			continue
+		}
+		lines = append(lines, line)
+	}
+	joined := strings.Join(lines, "\\n ")
```
kensipe, line 46:
```suggestion
	joined := strings.Join(lines, `\n `)
```
seems better to remove escape chars

kensipe (approved):
nothing blocking.. there is some concern of use.  initially I didn't realize the string created is just for display purposes.   It may be worth adding that in the godoc for `(c *Command) String() string {` 

The most concerning is the magic numbers.  It seems worth adding a constant having a name that clarifies purpose / intent.

It also likely makes sense to clarify the godoc on `summarize` 

It is all functional... as such I'll approve.  let me know when you are done and I'll merge if you prefer

pkg/apis/testharness/v1beta1/command.go
porridge: It's actually [this fmt.Stringer](https://pkg.go.dev/fmt#Stringer), not the stringer tool. Its docstring says:
> String method, which defines the “native” format for that value

unfortunately without explaining what "native format" means. :-(

I'm adding a docstring that clarifies the purpose.

pkg/apis/testharness/v1beta1/command.go
porridge: I don't intend the value returned by `String` to be machine-readable. I'm just trying to make sure that if for some reason in the future an invalid `Command` object is included in the `kuttl` log, then its description will be useful to a human. I feel like such message does this, unlike a simple empty string.

FTR, neither this nor the next case should ever happen in real life, since we already reject `Command` objects where `.Command == "" XOR .Script == ""` isn't met during unmarshalling. But better safe than sorry.

pkg/apis/testharness/v1beta1/command.go
porridge: Done

pkg/apis/testharness/v1beta1/command.go
porridge: Good point, docstring enhanced.

pkg/apis/testharness/v1beta1/command.go
porridge: Done

PR #496, opened 2023-11-01
upgrade controller-runtime to v0.16

### LINKED ISSUE #473: Update to controller-runtime v0.16+
**Why is this needed**: Big performance improvements for clusters with many CRDs installed via the new lazy restmapper client.

https://github.com/kubernetes-sigs/controller-runtime/pull/2116

There are some breaking changes though:

https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

davidebianchi:

Upgrade controller-runtime to version 16.

Fixes #473


pkg/test/step_integration_test.go
```diff
@@ -289,7 +289,8 @@ func TestStepDeleteExistingLabelMatch(t *testing.T) {
 	}), podSpec)
 
 	step := Step{
-		Logger: testutils.NewTestLogger(t, ""),
+		Logger:  testutils.NewTestLogger(t, ""),
+		Timeout: 60,
```
porridge, line 293: @davidebianchi I'm curious why this timeout field setting was needed? :thinking:

porridge (approved):

pkg/test/step_integration_test.go
davidebianchi: After the upgrade, this test will timeout. Adding this setting, it works correctly (without sensible increase in testing time).

porridge: Thanks. I wonder what changed. Have we been not enforcing a timeout (which I assume was 0 by default) before? :thinking:

davidebianchi: Mmh maybe it is related with this function change: https://github.com/kudobuilder/kuttl/pull/496/files#diff-f83daba5f9026600aa38b2b1e5a0c585535dc3355457e4762c50bdb0f6b0febdR160 (the PollImmediate function is deprecated). It could be that with 0, the context goes in timeout immediately.
Maybe we could change the implementation to use `PollUntilContextCancel` if timeout is 0. What do you think?

porridge: Ah, I think I know why. Previously [this Get](https://github.com/kudobuilder/kuttl/pull/496/files#diff-f83daba5f9026600aa38b2b1e5a0c585535dc3355457e4762c50bdb0f6b0febdR164) would run with an unbounded (TODO) context. Now it runs with a context with a 0 timeout.

I think the current code is fine, the real default timeout is not 0 in kuttl anyway.

PR #498, opened 2023-11-15
Elide uninteresting fields in object diffs.

### LINKED ISSUE #370: Filter `metadata.managedFields` from yaml diff output

Filter `metadata.managedFields` from yaml diff output when an expectation mismatch is printed. The fields should remain as is, if it's part of the expectation.

**Why is this needed**:

`metadata.managedFields`  contains a lot of information, that is seldom relevant. Printing it on every mismatch makes the output more verbose than needed. It hides the important differences and makes test results difficult to navigate.

### LINKED ISSUE #497: Suppress uninteresting parts of long object diffs
**What happened**:

Assert on something like a `Deployment`, for example:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: central
status:
  availableReplicas: 1
```

and the deployment's spec is non-trivial, you get this massive literally a few hundred lines long diff which mostly consists of `metadata.managedFields` (see #370) and contents of `spec` which is mostly irrelevant with the above assertion. This verbosity makes it somewhat hard to navigate the kuttl output, especially if you assert against several such deployments in single step.

The pretty-diff output should omit long unnecessary fields.

See the test case in the PR which is coming up soon.

porridge:

Makes diffs on large objects much less verbose but not any less useful, by omitting these fields in the object which are *not* specified in the assertion *and* are at least moderately long when serialized.

This way, asserting on `status.availableReplicas` of a Deployment (which is missing if zero replicas are available) will still show the `status.unavailableReplicas` for example, but will omit `spec` completely unless the assertion also mentions it.

This saves hundreds to thousands of lines of logs to scroll when debugging failures of my operator tests. See [the test](https://github.com/kudobuilder/kuttl/pull/498/files#diff-33f183dc21866dfa24678e9b67ddae498159fc33c737dd8167abe89326c70b4bR530-R568) for an example output. Before this change, the diff would contain the entirety of [the "actual" resource from a cluster](https://github.com/kudobuilder/kuttl/pull/498/files#diff-e1b4676c4b35129a792ba492478fdea244293c0a5d78284f983a2669dce58af0).

Fixes #497 
Fixes #370

### DIFF
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -378,9 +378,24 @@ func pruneLargeAdditions(expected *unstructured.Unstructured, actual *unstructur
 	return pruned
 }
 
+// prune replaces some fields in the actual tree to make it smaller for display.
+//
+// The goal is to make diffs on large objects much less verbose but not any less useful,
+// by omitting these fields in the object which are not specified in the assertion and are at least
+// moderately long when serialized.
+//
+// This way, for example when asserting on status.availableReplicas of a Deployment
+// (which is missing if zero replicas are available) will still show the status.unavailableReplicas
+// for example, but will omit spec completely unless the assertion also mentions it.
+//
+// This saves hundreds to thousands of lines of logs to scroll when debugging failures of some operator tests.
 func prune(expected map[string]interface{}, actual map[string]interface{}) {
+	// This value was chosen so that it is low enough to hide huge fields like `metadata.managedFields`,
+	// but large enough such that for example a typical `metadata.labels` still shows,
+	// since it might be useful for identifying reported objects like pods.
+	// This could potentially be turned into a knob in the future.
 	const maxLines = 10
-	toRemove := []string{}
+	var toRemove []string
 	for k, v := range actual {
 		if _, inExpected := expected[k]; inExpected {
 			expectedMap, isExpectedMap := expected[k].(map[string]interface{})
--- a/pkg/test/utils/kubernetes.go
+++ b/pkg/test/utils/kubernetes.go
@@ -372,8 +372,50 @@ func Namespaced(dClient discovery.DiscoveryInterface, obj runtime.Object, namesp
 	return m.GetName(), namespace, nil
 }
 
+func pruneLargeAdditions(expected *unstructured.Unstructured, actual *unstructured.Unstructured) runtime.Object {
+	pruned := actual.DeepCopy()
+	prune(expected.Object, pruned.Object)
+	return pruned
+}
+
+func prune(expected map[string]interface{}, actual map[string]interface{}) {
+	const maxLines = 10
+	toRemove := []string{}
+	for k, v := range actual {
+		if _, inExpected := expected[k]; inExpected {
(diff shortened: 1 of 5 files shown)


kovayur (approved):

pkg/test/utils/kubernetes.go
```diff
@@ -372,16 +372,58 @@ func Namespaced(dClient discovery.DiscoveryInterface, obj runtime.Object, namesp
...
 
+func pruneLargeAdditions(expected *unstructured.Unstructured, actual *unstructured.Unstructured) runtime.Object {
+	pruned := actual.DeepCopy()
+	prune(expected.Object, pruned.Object)
+	return pruned
+}
+
+func prune(expected map[string]interface{}, actual map[string]interface{}) {
+	const maxLines = 10
+	toRemove := []string{}
```
kensipe, line 383: might consider
```suggestion
	var toRemove []string
```

pkg/test/utils/kubernetes.go
```diff
@@ -372,16 +372,58 @@ func Namespaced(dClient discovery.DiscoveryInterface, obj runtime.Object, namesp
...
 }
 
+func pruneLargeAdditions(expected *unstructured.Unstructured, actual *unstructured.Unstructured) runtime.Object {
+	pruned := actual.DeepCopy()
+	prune(expected.Object, pruned.Object)
+	return pruned
+}
+
+func prune(expected map[string]interface{}, actual map[string]interface{}) {
+	const maxLines = 10
```
kensipe, line 382: how did we come to 10 as maxLines?

pkg/test/utils/kubernetes.go
```diff
@@ -372,16 +372,58 @@ func Namespaced(dClient discovery.DiscoveryInterface, obj runtime.Object, namesp
 	return m.GetName(), namespace, nil
 }
 
+func pruneLargeAdditions(expected *unstructured.Unstructured, actual *unstructured.Unstructured) runtime.Object {
+	pruned := actual.DeepCopy()
+	prune(expected.Object, pruned.Object)
+	return pruned
+}
+
+func prune(expected map[string]interface{}, actual map[string]interface{}) {
```
kensipe, line 381: seems worth a godoc to explain some of the logic or the purpose.  It's easy to understand based on name that we are "pruning" but pruning what and why and perhaps some how

kensipe (approved):
this is very welcomed!  nice work.  I left some suggestions but nothing blocking.   It's worth adding that `elided` is an uncommon word... that said, it seems appropriate.  

really like the end result and see lots of value to it!   great stuff!

kensipe:
re-reading the PR desc... the first paragraph or sentence would make a perfect godoc for the `prune` func

pkg/test/utils/kubernetes.go
porridge: I tried to find something that's low enough to hide huge fields like `metadata.managedFields`, but large enough such that typical `metadata.labels` still shows since it might be useful for identifying reported objects like pods. We can always turn this into a knob later...
Actually let me add such comment here.

PR #501, opened 2023-12-05
Make namespace auto-generated name more unique

### LINKED ISSUE #500: Auto-generated namespace name clashes
**What happened**:
When running Kuttl tests in CI, `--skip-delete` is used to speed up the test suite (~5x faster) as you don't have to wait for namespace deletion. A fresh cluster is used each CI run so leaving resources behind after a run is not an issue.

Periodically a namespace clash will occur where kuttl tries to create a new namespace when a namespace with the auto generated name already exists.

Example:
```
case.go:343: namespaces "kuttl-test-many-anteater" already exists
```

No namespace clash to occur

Run many test cases with `--skip-delete`

Y0sh1dk:
Bump the auto-generated namespace name words from 2 -> 4 to reduce the chance of a clash.

Fixes #500

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -385,7 +385,7 @@ func (t *Case) determineNamespace() *namespace {
 	}
 	// no preferred ns, means we auto-create with petnames
 	if t.PreferredNamespace == "" {
-		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
+		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(4, "-"))
 		ns.AutoCreated = true
 	}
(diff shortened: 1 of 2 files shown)


ajith-anz (approved):
Looks good

porridge (commented):
Thanks for the contribution, but I have a concern inline. Even though I despise adding yet another knob, I'm afraid depending on the answer to the question there, we might have to make this configurable, and keep the current 2 as default. :thinking:

porridge:
An alternative would be to use a different string of similar length as the current one, but containing more entropy. I.e. use a string of random alphanumeric characters rather than `petname`. The k8s code which generates unique pod or replicaset names for deployments might serve as an example...

PR #515, opened 2024-04-22
Replace inactive code owners with myself

porridge:

No point bothering inactive ones, while the notifications would be useful to myself.


erikgb (approved):

kaiwalyajoshi:
Thanks and godspeed @porridge @kensipe!

Issue #512, opened 2024-04-18
Specified test namespace not created when running TestSuite


Jaybee18:
**What happened**:
The tests in a TestSuite fail, because the specified namespace does not exist.

**What you expected to happen**:
The TestSuite should create the namespace that is specified in the YAML, as it is described in the documentation:
https://kuttl.dev/docs/testing/reference.html#testsuite:~:text=%22kuttl%2Dtest%22-,namespace,-string

**How to reproduce it (as minimally and precisely as possible)**:
```yaml
# kuttl-test.yaml
apiVersion: kuttl.dev/v1beta1
kind: TestSuite
testDirs:
- ./mongodb
parallel: 1
namespace: "kuttl-tests-namespace"
```

```yaml
# mongodb/01-install.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo
data:
  # property-like keys; each key maps to a simple value
  player_initial_lives: "3"
  ui_properties_file_name: "user-interface.properties"

  # file-like keys
  game.properties: |
    enemy.types=aliens,monsters
    player.maximum-lives=5    
  user-interface.properties: |
    color.good=purple
    color.bad=yellow
    allow.textmode=true  
```

And then run
```sh
kubectl kuttl test .
```

**Anything else we need to know?**:
The logs indicate that the namespace creation was purposefully skipped:
```
logger.go:42: 09:30:26 | mongodb | Skipping creation of user-supplied namespace: kuttl-tests-namespace
```
**Environment**:
- Kubernetes version (use `kubectl version`): v1.27.11-eks-b9c9ed7
- KUTTL version (use `kubectl kuttl version`):
KUTTL Version: version.Info{GitVersion:"0.15.0", GitCommit:"f6d64c9", BuildDate:"2023-01-03T21:32:35Z", GoVersion:"go1.19.4", Compiler:"gc", Platform:"darwin/arm64"}
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #518, opened 2024-05-08
fix: create namespace when running TestSuite

chenmj11:

The tests in a TestSuite fail, because the specified namespace does not exist.

kuttl-test.yaml
```
apiVersion: kuttl.dev/v1beta1
kind: TestSuite
namespace: test
timeout: 60
reportName: test
reportFormat: JSON 
```
error logs
```
    logger.go:42: 13:48:07 | pod | Ignoring pod-assert.yaml as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$  
    logger.go:42: 13:48:07 | pod | Ignoring pod.yaml as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$  
    logger.go:42: 13:48:07 | pod | Skipping creation of user-supplied namespace: test  
    logger.go:42: 13:48:07 | pod/0-step | starting test step 0-step  
    case.go:366: failed in step 0-step  
    case.go:368: namespaces "test" not found  
```
*Automatically closes linked issue when PR is merged.
Fixes [512](https://github.com/kudobuilder/kuttl/issues/512)

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -316,7 +316,12 @@ func shortString(obj *corev1.ObjectReference) string {
 // Run runs a test case including all of its steps.
 func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 	setupReport := report.NewCase("setup")
-	ns := t.determineNamespace()
+	ns, err := t.determineNamespace()
+	if err != nil {
+		setupReport.Failure = report.NewFailure(err.Error(), nil)
+		ts.AddTestcase(setupReport)
+		test.Fatal(err)
+	}
 
 	cl, err := t.Client(false)
 	if err != nil {
@@ -388,7 +393,7 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 	}
 }
 
-func (t *Case) determineNamespace() *namespace {
+func (t *Case) determineNamespace() (*namespace, error) {
 	ns := &namespace{
 		Name:        t.PreferredNamespace,
 		AutoCreated: false,
@@ -397,9 +402,17 @@ func (t *Case) determineNamespace() *namespace {
 	if t.PreferredNamespace == "" {
 		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
 		ns.AutoCreated = true
+	} else {
+		exist, err := t.NamespaceExists(t.PreferredNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine existence of namespace %q: %w", t.PreferredNamespace, err)
+		}
+		if !exist {
+			ns.AutoCreated = true
+		}
 	}
-	// if we have a preferred namespace, we do NOT auto-create
-	return ns
+	// if we have a preferred namespace, and it already exists, we do NOT auto-create
+	return ns, nil
 }
 
 // CollectTestStepFiles collects a map of test steps and their associated files


pkg/test/case.go
```diff
@@ -397,6 +397,11 @@ func (t *Case) determineNamespace() *namespace {
 	if t.PreferredNamespace == "" {
 		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
 		ns.AutoCreated = true
+	} else {
+		exist, _ := t.NamespaceExists(t.PreferredNamespace)
```
porridge, line 401: Please do propagate and fail the test if the check fails due to bad cluster config, etc.

porridge (changes requested):
Thank you for the contribution!
I think this looks good apart from the hiding of error.

pkg/test/case.go
```diff
@@ -397,6 +397,14 @@ func (t *Case) determineNamespace() *namespace {
 	if t.PreferredNamespace == "" {
 		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
 		ns.AutoCreated = true
+	} else {
+		exist, err := t.NamespaceExists(t.PreferredNamespace)
+		if err != nil {
+			t.Logger.Logf("found namespace %s err, %s", t.PreferredNamespace, err.Error())
```
porridge, line 403: I don't think we want to log+ignore an error when retrieving namespace. If we're not able to even look up its existence then creating it is even less likely to succeed. How about changing `determineNamespace` to return `(*namespace, error)` and then failing the run if `err != nil` in its caller(s)?

porridge (changes requested):

pkg/test/case.go
chenmj11: yes, I aggree. This looks much better.

pkg/test/case.go
```diff
@@ -397,9 +402,17 @@ func (t *Case) determineNamespace() *namespace {
...
+	} else {
+		exist, err := t.NamespaceExists(t.PreferredNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("found namespace %s err, %s", t.PreferredNamespace, err.Error())
+		}
+		if !exist {
+			ns.AutoCreated = true
+		}
 	}
 	// if we have a preferred namespace, we do NOT auto-create
```
porridge, line 414: This comment is now stale, and in the wrong place...
Please move it into the else block, and change to say something like:
```suggestion
	// if we have a preferred namespace, and it already exists, we do NOT auto-create
```

pkg/test/case.go
```diff
@@ -397,9 +402,17 @@ func (t *Case) determineNamespace() *namespace {
 	if t.PreferredNamespace == "" {
```
porridge, line 402: If the latter comment will be inside the else block, please move the one above into the if block for consistency.

pkg/test/case.go
```diff
@@ -397,9 +402,17 @@ func (t *Case) determineNamespace() *namespace {
...
 		ns.Name = fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-"))
 		ns.AutoCreated = true
+	} else {
+		exist, err := t.NamespaceExists(t.PreferredNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("found namespace %s err, %s", t.PreferredNamespace, err.Error())
```
porridge, line 408: I think something like this would be more fitting and robust to weird ns names:
```suggestion
			return nil, fmt.Errorf("failed to determine existence of namespace %q: %w", t.PreferredNamespace, err)
```

porridge (changes requested):
Nice! Just a few more nitpicks and this will be good to merge.

pkg/test/case.go
chenmj11: The comments are all outside

porridge:
Looking closer at this, the `determineNamespace()` call should probably be made using all clients once their list is populated a bit later. Otherwise we risk creating the ns on some clusters but not others, in case we get an error because the namespace already exists on some of them.

However I'm going to merge this now since it's already an improvement in the common single-cluster case.

porridge (approved):
Thank you for the contribution @chenmj11 !

PR #529, opened 2024-06-10
Bump dependencies and group them.

porridge:

- Create a single PR for all bumps created by dependabot since they typically need to go in lockstep.
- Also needed to update codegen scripts since [the old script we used](https://github.com/kubernetes/code-generator/blob/8d055c77cf9e3982f9ed9c46cb6c614eafdc18ff/generate-groups.sh#L17-L18) is now EOL.


hack/update_codegen.sh
```diff
@@ -21,18 +23,10 @@ else
...
 
-# Fake being in a $GOPATH until kubernetes fully supports modules
-# https://github.com/kudobuilder/kudo/issues/1252
-FAKE_GOPATH="$(mktemp -d)"
-trap 'chmod -R u+rwX ${FAKE_GOPATH} && rm -rf ${FAKE_GOPATH}' EXIT
-FAKE_REPOPATH="${FAKE_GOPATH}/src/github.com/kudobuilder/kuttl"
-mkdir -p "$(dirname "${FAKE_REPOPATH}")" && ln -s "${REPO_ROOT}" "${FAKE_REPOPATH}"
-export GOPATH="${FAKE_GOPATH}"
-cd "${FAKE_REPOPATH}"
+# Set GOBIN make gen_helpers install and run binaries in the versioned directory.
```
mclasmeier, line 26:
```suggestion
# Set GOBIN to make gen_helpers install and run binaries in the versioned directory.
```

mclasmeier (approved):

PR #540, opened 2024-07-03
feat: add lazy loading kubeconfigs for Test Steps

### LINKED ISSUE #539: Support lazy-loading of Kubeconfig in the Test Steps
Enable Kuttl to load the Kubeconfig for the test-step when the test runs.

**Why is this needed**:
Currently, Kuttl creates clients for each Test Step when the Test Case starts running. Although this allows early detection of erroneous Kubeconfigs in the Steps, this prevents lazy-loading of Kubeconfigs.

We have a use-case where the Kubeconfig for a Test Step is "created by an earlier Test Step." Because the Kubeconfig doesn't exist before the older Test Step is completed, the Test Case itself fails.

kumar-mallikarjuna:
This PR adds an attribute `LazyLoadKubeconfig` to the `TestStep` type which can be used to configure whether the `Kubeconfig` specified in the `TestStep` needs to be lazy-loaded, i.e., if the `Kubeconfig` doesn't need to be loaded at the start of the `TestCase` eval.

Fixes #539

**Tests**
```
.
├── eager
│   └── loading
│       └── 00-teststep.yaml
├── kubeconfig
├── kuttl-test.yaml
└── lazy
    └── loading
        └── 00-teststep.yaml
```

`eager/loading/00-teststep.yaml`:
```yaml
apiVersion: kuttl.dev/v1beta1
kind: TestStep
commands:
- script: echo hi
kubeconfig: test.kubeconfig
kubeconfigLoading: Eager
```

`lazy/loading/00-teststep.yaml`:
```yaml
apiVersion: kuttl.dev/v1beta1
kind: TestStep
commands:
- script: echo hi
kubeconfig: test.kubeconfig
kubeconfigLoading: Lazy
```

Output on `main`:
```shell
➜  lazy_test git:(main) ✗ go run ../cmd/kubectl-kuttl/main.go test eager
2024/07/04 11:43:22 kutt-test config testdirs is overridden with args: [ eager ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:49639
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: eager has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    case.go:344: stat /Users/kumarmallikarjuna/Workspace/oss/kuttl/lazy_test/eager/loading/test.kubeconfig: no such file or directory
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (0.01s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/loading (0.00s)
FAIL
exit status 1
➜  lazy_test git:(main) ✗ go run ../cmd/kubectl-kuttl/main.go test lazy
2024/07/04 11:43:27 kutt-test config testdirs is overridden with args: [ lazy ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:49639
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: lazy has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    case.go:344: stat /Users/kumarmallikarjuna/Workspace/oss/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (0.02s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/loading (0.00s)
FAIL
exit status 1
```

Output after these changes:
```shell
➜  lazy_test git:(tmp-kubeconfig-lazy-loading) ✗ go run ../cmd/kubectl-kuttl/main.go test eager
2024/07/04 11:41:15 kutt-test config testdirs is overridden with args: [ eager ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:6443
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: eager has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    case.go:346: stat /Users/kumarmallikarjuna/Workspace/oss/kuttl/lazy_test/eager/loading/test.kubeconfig: no such file or directory
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (0.01s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/loading (0.00s)
FAIL
exit status 1
➜  lazy_test git:(tmp-kubeconfig-lazy-loading) ✗ go run ../cmd/kubectl-kuttl/main.go test lazy
2024/07/04 11:41:18 kutt-test config testdirs is overridden with args: [ lazy ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:6443
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: lazy has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    logger.go:42: 11:41:18 | loading | Creating namespace: kuttl-test-credible-crow
    case.go:400: failed in step 0-teststep
    case.go:402: failed to lazy-load kubeconfig: stat /Users/kumarmallikarjuna/Workspace/oss/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
    logger.go:42: 11:41:18 | loading | loading events from ns kuttl-test-credible-crow:
    logger.go:42: 11:41:18 | loading | Deleting namespace: kuttl-test-credible-crow
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (5.14s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/loading (5.12s)
FAIL
exit status 1
```

Lazy-loading when the Kubeconfig exists and is same as the global Kubeconfig:
```shell
➜  lazy_test git:(main) ✗ go run ../cmd/kubectl-kuttl/main.go test eager
2024/07/04 11:37:44 kutt-test config testdirs is overridden with args: [ eager ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:49639
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: eager has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    case.go:344: stat /Users/kumarmallikarjuna/Workspace/oss/kuttl/lazy_test/eager/loading/test.kubeconfig: no such file or directory
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (0.02s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/loading (0.00s)
FAIL
exit status 1
➜  lazy_test git:(main) ✗ go run ../cmd/kubectl-kuttl/main.go test lazy
2024/07/04 11:37:50 kutt-test config testdirs is overridden with args: [ lazy ]
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at: https://127.0.0.1:49639
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 2400 seconds for each step
    harness.go:375: testsuite: lazy has 1 tests
=== RUN   kuttl/harness
=== RUN   kuttl/harness/loading
=== PAUSE kuttl/harness/loading
=== CONT  kuttl/harness/loading
    logger.go:42: 11:37:50 | loading | Ignoring test.kubeconfig as it does not match file name regexp: ^(\d+)-(?:[^\.]+)(?:\.yaml)?$
    logger.go:42: 11:37:50 | loading | Creating namespace: kuttl-test-needed-ox
    logger.go:42: 11:37:50 | loading | Creating namespace: kuttl-test-needed-ox
    logger.go:42: 11:37:50 | loading/0-teststep | starting test step 0-teststep
    logger.go:42: 11:37:50 | loading/0-teststep | running command: [sh -c echo hi]
    logger.go:42: 11:37:50 | loading/0-teststep | hi
    logger.go:42: 11:37:50 | loading/0-teststep | test step completed 0-teststep
    logger.go:42: 11:37:50 | loading | loading events from ns kuttl-test-needed-ox:
    logger.go:42: 11:37:50 | loading | Deleting namespace: kuttl-test-needed-ox
    logger.go:42: 11:37:55 | loading | Deleting namespace: kuttl-test-needed-ox
```

### DIFF
--- a/docs/testing/reference.md
+++ b/docs/testing/reference.md
@@ -73,7 +73,7 @@ delete   | list of object references | A list of objects to delete, if they do n
 index    | int                       | Override the test step's index.
 commands | list of [Commands](#commands) | Commands to run prior at the beginning of the test step.
 kubeconfig    | string                       | The Kubeconfig file to use to run the included steps(s).
-kubeconfigLoading    | string                       | Specifies the mode for loading Kubeconfig: Eager/Lazy. Defaults to Eager.
+kubeconfigLoading    | string                | Specifies the mode for loading Kubeconfig and making a cluster connection: `Eager` (when loading the test definition) or `Lazy` (right before executing the step, makes it possible to generate the Kubeconfig in a preceding step). Defaults to `Eager`.
 unitTest    | bool                       | Indicates if the step is a unit test, safe to run without a real Kubernetes cluster.
 
 
--- a/docs/testing/reference.md
+++ b/docs/testing/reference.md
@@ -73,6 +73,7 @@ delete   | list of object references | A list of objects to delete, if they do n
 index    | int                       | Override the test step's index.
 commands | list of [Commands](#commands) | Commands to run prior at the beginning of the test step.
 kubeconfig    | string                       | The Kubeconfig file to use to run the included steps(s).
+kubeconfigLoading    | string                       | Specifies the mode for loading Kubeconfig: Eager/Lazy. Defaults to Eager.
 unitTest    | bool                       | Indicates if the step is a unit test, safe to run without a real Kubernetes cluster.
 
 
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -6,6 +6,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const KubeconfigLoadingEager = "Eager"
 const KubeconfigLoadingLazy = "Lazy"
 
 // Create embedded struct to implement custom DeepCopyInto method
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -6,6 +6,8 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const KubeconfigLoadingLazy = "Lazy"
+
 // Create embedded struct to implement custom DeepCopyInto method
 type RestConfig struct {
 	RC *rest.Config
@@ -126,8 +128,10 @@ type TestStep struct {
 	// Kubeconfig to use when applying and asserting for this step.
 	Kubeconfig string `json:"kubeconfig,omitempty"`
 
-	// Loads kubeconfig only when the previous steps have finished.
-	LazyLoadKubeconfig bool `json:"lazyLoadKubeconfig,omitempty"`
+	// Specifies the mode for loading Kubeconfig: Eager/Lazy. Defaults to Eager.
+	// +kubebuilder:default=Eager
+	// +kubebuilder:validation:Enum=Eager;Lazy
+	KubeconfigLoading string `json:"kubeconfigLoading,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/testharness/v1beta1/test_types.go
+++ b/pkg/apis/testharness/v1beta1/test_types.go
@@ -125,6 +125,9 @@ type TestStep struct {
 
 	// Kubeconfig to use when applying and asserting for this step.
 	Kubeconfig string `json:"kubeconfig,omitempty"`
+
+	// Loads kubeconfig only when the previous steps have finished.
+	LazyLoadKubeconfig bool `json:"lazyLoadKubeconfig,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -376,7 +376,7 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 		if testStep.KubeconfigLoading == v1beta1.KubeconfigLoadingLazy {
 			cl, err = newClient(testStep.Kubeconfig)(false)
 			if err != nil {
-				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig '%v': %w", testStep.Kubeconfig, err))
+				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig %q: %w", testStep.Kubeconfig, err))
 			} else {
 				clients[testStep.Kubeconfig] = cl
 			}
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -373,6 +373,7 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 
 		errs := []error{}
 
+		// Set-up client for lazy-loaded Kubeconfigs
 		if testStep.KubeconfigLoading == v1beta1.KubeconfigLoadingLazy {
 			cl, err = newClient(testStep.Kubeconfig)(false)
 			if err != nil {
@@ -382,7 +383,11 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 			}
 		}
 
-		errs = append(errs, testStep.Run(test, ns.Name)...)
+		// Run test case only if no setup errors are encountered
+		if len(errs) == 0 {
+			errs = append(errs, testStep.Run(test, ns.Name)...)
+		}
+
 		if len(errs) > 0 {
 			caseErr := fmt.Errorf("failed in step %s", testStep.String())
 			tc.Failure = report.NewFailure(caseErr.Error(), errs)
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
 	eventsbeta1 "k8s.io/api/events/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -348,8 +349,10 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 		clients[testStep.Kubeconfig] = cl
 	}
 
-	for _, c := range clients {
-		if err = t.CreateNamespace(test, c, ns); err != nil {
+	for kc, c := range clients {
+		if err = t.CreateNamespace(test, c, ns); apierrors.IsAlreadyExists(err) {
+			t.Logger.Logf("namespace %q already exists, using kubeconfig %q", ns.Name, kc)
+		} else if err != nil {
 			setupReport.Failure = report.NewFailure("failed to create test namespace", []error{err})
 			ts.AddTestcase(setupReport)
 			test.Fatal(err)
@@ -378,7 +381,9 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 			cl, err = testStep.Client(false)
 			if err != nil {
 				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig: %w", err))
-			} else if err = t.CreateNamespace(test, cl, ns); err != nil {
+			} else if err = t.CreateNamespace(test, cl, ns); apierrors.IsAlreadyExists(err) {
+				t.Logger.Logf("namespace %q already exists", ns.Name)
+			} else if err != nil {
 				errs = append(errs, fmt.Errorf("failed to create test namespace: %w", err))
 			}
 		}
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -16,7 +16,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
 	eventsbeta1 "k8s.io/api/events/v1beta1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -350,7 +349,7 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 	}
 
 	for kc, c := range clients {
-		if err = t.CreateNamespace(test, c, ns); apierrors.IsAlreadyExists(err) {
+		if err = t.CreateNamespace(test, c, ns); k8serrors.IsAlreadyExists(err) {
 			t.Logger.Logf("namespace %q already exists, using kubeconfig %q", ns.Name, kc)
 		} else if err != nil {
 			setupReport.Failure = report.NewFailure("failed to create test namespace", []error{err})
@@ -381,7 +380,7 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 			cl, err = testStep.Client(false)
 			if err != nil {
 				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig: %w", err))
-			} else if err = t.CreateNamespace(test, cl, ns); apierrors.IsAlreadyExists(err) {
+			} else if err = t.CreateNamespace(test, cl, ns); k8serrors.IsAlreadyExists(err) {
 				t.Logger.Logf("namespace %q already exists", ns.Name)
 			} else if err != nil {
 				errs = append(errs, fmt.Errorf("failed to create test namespace: %w", err))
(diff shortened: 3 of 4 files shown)


pkg/test/case.go
```diff
@@ -370,6 +370,16 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 		tc.Assertions += len(testStep.Asserts)
 		tc.Assertions += len(testStep.Errors)
 
+		if testStep.LazyLoadKubeConfig {
+			cl, err = newClient(testStep.Kubeconfig)(false)
+			if err != nil {
+				setupReport.Failure = report.NewFailure(err.Error(), nil)
+				ts.AddTestcase(setupReport)
+				test.Fatal(err)
```
porridge, line 378: At this point the `setupReport` is already added (line 357).
I think on error here we should instead do something similar to what happens in lines 385-390.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -125,6 +125,9 @@ type TestStep struct {
 
 	// Kubeconfig to use when applying and asserting for this step.
 	Kubeconfig string `json:"kubeconfig,omitempty"`
+
+	// Loads kubeconfig only when the previous steps have finished.
+	LazyLoadKubeconfig bool `json:"lazyLoadKubeconfig,omitempty"`
```
porridge, line 130: Please read the comment on bools in [this conventions doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) and turn this field into something like `kubeConfigLoading` with allowed values of `Eager` and `Lazy`.
Also please update the `./docs/` in this repo.

porridge (changes requested):
Thanks for the contribution!
Please see the comments inline.

Also, could you please either add an automated test, or if (likely) this will be difficult, then paste how the kuttl output looks like in case the lazy loading fails (perhaps in comparison to how it fails eagerly now).

kumar-mallikarjuna:
Hi @porridge , thanks for the quick review. I've incorporated your comments. I found it a little difficult to add tests for this. For now, I've updated the PR description with the outputs comparing the failures to load lazily on `main` and with these changes. Can you please take a look again?

kumar-mallikarjuna:
I see the CRDS are in `crds/`. But I'm unable to generate them with `make generate`. Is this expected?

pkg/test/case.go
```diff
@@ -370,7 +371,18 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 		tc.Assertions += len(testStep.Asserts)
 		tc.Assertions += len(testStep.Errors)
 
-		errs := testStep.Run(test, ns.Name)
+		errs := []error{}
+
+		if testStep.KubeconfigLoading == v1beta1.KubeconfigLoadingLazy {
+			cl, err = newClient(testStep.Kubeconfig)(false)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig '%v': %w", testStep.Kubeconfig, err))
```
porridge, line 379:
```suggestion
				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig %q: %w", testStep.Kubeconfig, err))
```

pkg/test/step.go
```diff
@@ -555,6 +556,7 @@ func (s *Step) LoadYAML(file string) error {
 				exKubeconfig := env.Expand(s.Step.Kubeconfig)
 				s.Kubeconfig = cleanPath(exKubeconfig, s.Dir)
 			}
+			s.KubeconfigLoading = s.Step.KubeconfigLoading
```
porridge, line 559: I'm not sure if the kubebuilder markers are respected.
Can you please validate somewhere here that the value is either one of the two allowed values or empty?

docs/testing/reference.md
```diff
@@ -73,6 +73,7 @@ delete   | list of object references | A list of objects to delete, if they do n
 index    | int                       | Override the test step's index.
 commands | list of [Commands](#commands) | Commands to run prior at the beginning of the test step.
 kubeconfig    | string                       | The Kubeconfig file to use to run the included steps(s).
+kubeconfigLoading    | string                       | Specifies the mode for loading Kubeconfig: Eager/Lazy. Defaults to Eager.
```
porridge, line 76:
```suggestion
kubeconfigLoading    | string                       | Specifies the mode for loading Kubeconfig and making a cluster connection: `Eager` (when loading the test definition) or `Lazy` (right before executing the step, makes it possible to generate the Kubeconfig in a preceding step). Defaults to `Eager`.
```

porridge (changes requested):
Nice!

I don't see the `failed to lazy-load kubeconfig` string in the ouput you pasted? Was the code changed after you copied? Can you please update them?

As for the CRDs, they seem to have gone stale. I didn't even know they were there 😅 

If you could file a bug to remind us to fix generation and check their freshness in CI, that would be great.

This is almost good to go, just a few more nitpicks.

kumar-mallikarjuna:
> I don't see the failed to lazy-load kubeconfig string in the ouput you pasted? Was the code changed after you copied? Can you please update them?

Apologies, I see to have mixed up the outputs. I've updated them now.

> As for the CRDs, they seem to have gone stale. I didn't even know they were there 😅

Oh, okay.

> If you could file a bug to remind us to fix generation and check their freshness in CI, that would be great.

Opened https://github.com/kudobuilder/kuttl/issues/541.

pkg/test/case.go
```diff
@@ -370,7 +371,18 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
...
+		if testStep.KubeconfigLoading == v1beta1.KubeconfigLoadingLazy {
+			cl, err = newClient(testStep.Kubeconfig)(false)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("failed to lazy-load kubeconfig %q: %w", testStep.Kubeconfig, err))
+			} else {
+				clients[testStep.Kubeconfig] = cl
+			}
+		}
+
+		errs = append(errs, testStep.Run(test, ns.Name)...)
```
porridge, line 385: I noticed in the output that we're getting the error twice now:

```
    case.go:392: failed to lazy-load kubeconfig "/Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig": stat /Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
    case.go:392: stat /Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
```

I guess this is because we `Run` the step on this line regardless of whether the lazy loading succeeded or not 😟 

However looking deeper into this I realized a few more problems that I didn't see earlier, sorry:
- the conditional creation of a client on line 377 is mostly redundant since we already prepare one on line 364,
- more importantly, we're failing to create a namespace for lazy-loaded clients, I think we should replace the block  377-382 with one that creates a namespace if it hasn't been done eagerly (not sure how to do this elegantly and robustly, but I'm sure you can come up with something 😉 )

porridge (changes requested):

pkg/test/case.go
kumar-mallikarjuna: > I noticed in the output that we're getting the error twice now:
> 
> ```
>     case.go:392: failed to lazy-load kubeconfig "/Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig": stat /Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
>     case.go:392: stat /Users/kumarmallikarjuna/Workspace/kuttl/lazy_test/lazy/loading/test.kubeconfig: no such file or directory
> ```
> 
> I guess this is because we `Run` the step on this line regardless of whether the lazy loading succeeded or not 😟

Fixed here: https://github.com/kudobuilder/kuttl/pull/540/commits/75eb6a9b3fc25634b10b2b7c9d5022e2d9b16a1d. I've skipped the test setup from running in-case the kubeconfig setup fails.

> However looking deeper into this I realized a few more problems that I didn't see earlier, sorry:
> 
> * the conditional creation of a client on line 377 is mostly redundant since we already prepare one on line 364,
> * more importantly, we're failing to create a namespace for lazy-loaded clients, I think we should replace the block  377-382 with one that creates a namespace if it hasn't been done eagerly (not sure how to do this elegantly and robustly, but I'm sure you can come up with something 😉 )

Totally missed that. Since, the namespace is created for each eager client, I've replicated that and placed the logic right after the lazy-client is created:
https://github.com/kudobuilder/kuttl/pull/540/commits/82fe683e5317454767fced3e51f517d1232164cd.

kumar-mallikarjuna: Possible edge case here: the Kubeconfigs are common b/w steps or the same as the global one. In that case, the namespace would attempt to be re-created and its failure would keep the test from evaluating. I'll add an additional check.

pkg/test/case.go
```diff
@@ -347,9 +349,11 @@ func (t *Case) Run(test *testing.T, ts *report.Testsuite) {
 		clients[testStep.Kubeconfig] = cl
 	}
 
-	for _, c := range clients {
-		if err := t.CreateNamespace(test, c, ns); err != nil {
-			setupReport.Failure = report.NewFailure(err.Error(), nil)
+	for kc, c := range clients {
+		if err = t.CreateNamespace(test, c, ns); apierrors.IsAlreadyExists(err) {
+			t.Logger.Logf("namespace %q already exists, using kubeconfig %q", ns.Name, kc)
```
porridge, line 354: I'm having mixed feelings about this.

Before your change, encountering an existing namespace when trying to create one would abort the test and leave namespace alone.
After your change, it will silently assume ownership, and later delete the namespace.

The new behaviour is probably the expected one in a typical case, but could be a disaster if the user didn't mean to reuse (nor delete) the namespace.

I feel it should at least be documented.

WDYT?

porridge:
Also please take a look at the failing checks.

pkg/test/case.go
kumar-mallikarjuna: > After your change, it will silently assume ownership, and later delete the namespace.

Is there an ownership label that Kuttl adds to its resources? Ideally, the namespaces (and other resources) Kuttl is creating should only be cleaned if they have that label.

porridge: That's a good idea. But I don't know the answer. I think I've seen something like that somewhere in the code, but not sure. Shouldn't be hard to find out experimentally, though. And if there is no such thing, we can add this.

porridge: But probably in a different PR :-)

kumar-mallikarjuna: > Before your change, encountering an existing namespace when trying to create one would abort the test and leave namespace alone. After your change, it will silently assume ownership, and later delete the namespace.

Looks like `CreateNamespace` is only called if the Namespace is supposed to be auto-created: https://github.com/kudobuilder/kuttl/blob/280792ae8ad0b4fb1b4387a2ef085ac8daabdf3f/pkg/test/case.go#L100-L103
and so Namespaces not created by Kuttl shouldn't be cleaned up:
https://github.com/kudobuilder/kuttl/blob/280792ae8ad0b4fb1b4387a2ef085ac8daabdf3f/pkg/test/case.go#L113-L119

> The new behaviour is probably the expected one in a typical case, but could be a disaster if the user didn't mean to reuse (nor delete) the namespace.
> 
> I feel it should at least be documented.

I think `CreateNamespace` ensures that we're not deleting Namespaces Kuttl didn't create and if the user wants to keep it, they can specify `SkipDelete` in the `TestSuite`.

porridge: Ah, nice catch!

porridge (approved):

Issue #547, opened 2024-08-05
Allow user impersonation as a cli flag


logan-bobo:
**What would you like to be added**: The ability to pass an impersonation flag to `kuttl` to allow for the [impersonation of users](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation)

**Why is this needed**: There are situations where you may only have access to a single user that is able to impersonate other users that have privileged access. This would allow for the `--as` flag to be passed to `kuttl` to enable user impersonation. For example `kubectl kuttl test ... --as <user>`

Happy to pick this up if accepted ✋

porridge:
I hope there is a library we can use to be compatible with kubectl w.r.t. translation of the command line flag value to the actual headers that the client needs to send? 

Otherwise (if we do the translation by hand) it would be a shame to become subtly incompatible over time.

porridge:
Sounds good! It would be awesome if you could contribute this feature!

logan-bobo:
@porridge I believe we can also now close this one out.

porridge:
Done in https://github.com/kudobuilder/kuttl/pull/549

PR #549, opened 2024-08-05
feat: add user impersonation to all commands 

logan-bobo:

Fixes # https://github.com/kudobuilder/kuttl/issues/547

Working configuration after adding the `--as` flag for user impersonation.

```
./bin/kubectl-kuttl test --config ... --as admin
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at:...
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 30 seconds for each step
    harness.go:375: testsuite: ...
=== RUN   kuttl/harness
=== RUN   kuttl/harness/00-check-deploy
=== PAUSE kuttl/harness/00-check-deploy
=== CONT  kuttl/harness/00-check-deploy
    logger.go:42: 19:45:28 | 00-check-deploy | Creating namespace: kuttl-test-engaged-cockatoo
    logger.go:42: 19:45:28 | 00-check-deploy/0- | starting test step 0-
    logger.go:42: 19:45:29 | 00-check-deploy/0- | test step completed 0-
    logger.go:42: 19:45:29 | 00-check-deploy | skipping kubernetes event logging
    logger.go:42: 19:45:29 | 00-check-deploy | Deleting namespace: kuttl-test-engaged-cockatoo
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- PASS: kuttl (7.71s)
        --- PASS: kuttl/harness/00-check-deploy (7.16s)
PASS
```

Without having user impersonation

```
❯ ./bin/kubectl-kuttl test --config ...
=== RUN   kuttl
    harness.go:464: starting setup
    harness.go:255: running tests using configured kubeconfig.
    harness.go:278: Successful connection to cluster at:...
    harness.go:363: running tests
    harness.go:75: going to run test suite with timeout of 30 seconds for each step
    harness.go:375: testsuite:...
=== RUN   kuttl/harness
=== RUN   kuttl/harness/00-check-deploy
=== PAUSE kuttl/harness/00-check-deploy
=== CONT  kuttl/harness/00-check-deploy
    logger.go:42: 19:47:44 | 00-check-deploy | Creating namespace: kuttl-test-awaited-dingo
    case.go:357: namespaces is forbidden: User "" cannot create resource "namespaces" in API group "" at the cluster scope
    logger.go:42: 19:47:44 | 00-check-deploy | Deleting namespace: kuttl-test-awaited-dingo
    case.go:117: namespaces "kuttl-test-awaited-dingo" is forbidden: User "" cannot delete resource "namespaces" in API group "" in the namespace "kuttl-test-awaited-dingo"
=== NAME  kuttl
    harness.go:407: run tests finished
    harness.go:515: cleaning up
    harness.go:572: removing temp folder: ""
--- FAIL: kuttl (0.84s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/00-check-deploy (0.31s)
FAIL
```


pkg/kuttlctl/cmd/root.go
```diff
@@ -34,6 +37,7 @@ and serves as an API aggregation layer.
 	cmd.AddCommand(newErrorsCmd())
 	cmd.AddCommand(newTestCmd())
 	cmd.AddCommand(newVersionCmd())
+	configFlags.AddFlags(cmd.PersistentFlags())
```
logan-bobo, line 40: With this the following is added to the default command

```
❯ bin/kubectl-kuttl --help
...
Flags:
      --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
      --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --as-uid string                  UID to impersonate for the operation.
      --cache-dir string               Default cache directory (default "/Users/lcx/.kube/cache")
      --certificate-authority string   Path to a cert file for the certificate authority
      --client-certificate string      Path to a client certificate file for TLS
      --client-key string              Path to a client key file for TLS
      --cluster string                 The name of the kubeconfig cluster to use
      --context string                 The name of the kubeconfig context to use
      --disable-compression            If true, opt-out of response compression for all requests to the server
  -h, --help                           help for kubectl-kuttl
      --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kubeconfig string              Path to the kubeconfig file to use for CLI requests.
  -n, --namespace string               If present, the namespace scope for this CLI request
      --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
  -s, --server string                  The address and port of the Kubernetes API server
      --tls-server-name string         Server name to use for server certificate validation. If it is not provided, the hostname used to contact the server is used
      --token string                   Bearer token for authentication to the API server
      --user string                    The name of the kubeconfig user to use

Use "kubectl-kuttl [command] --help" for more information about a command.
...
```

A rest client config can be generated from these flags https://pkg.go.dev/k8s.io/cli-runtime/pkg/genericclioptions#ConfigFlags.ToRESTConfig

porridge: Hmm, most od these seem specific enough. It would be great to achieve this level of consistency with `kubectl`.

However seeing `namespace` there got me thinking... `kuttl test` subcommand at least [already has a namespace flag](https://github.com/kudobuilder/kuttl/blob/762b0f9d2a6d64816ba80bf41b39d97fbcf8c215/pkg/kuttlctl/cmd/test.go#L260). I wonder how the two definitions would interact... 🤔 

And related to that: how do we make sure nothing breaks (or at least we are able to notice before releasing) in the future when we pick up a new version of the library which adds a new flag which we already happen to be using for another purpose 😟 

Thoughts?

porridge: Ah, if I understand it correctly, this library [just copies](https://github.com/kubernetes/cli-runtime/blob/dd30b60637c4ce4981151982b7437d63663da7ec/pkg/genericclioptions/config_flags.go#L181) the value of the flag into the rest config. If it's that simple, then maybe we should do just that for the three `--as*` flags, to be on the safe side?

logan-bobo: Yeah I agree adding all of the flags from `genericclioptions` might be too wide a scope for this change and have the potential to collide with kuttl functionality so would take more time and care to implement along with test. 

> Ah, if I understand it correctly, this library [just copies](https://github.com/kubernetes/cli-runtime/blob/dd30b60637c4ce4981151982b7437d63663da7ec/pkg/genericclioptions/config_flags.go#L181) the value of the flag into the rest config. The heavy lifting is done by client-go? If it's that simple, then maybe we should do just that for the three `--as*` flags, to be on the safe side?

Lets do it! I will take this PR in the direction of adding the three `--as*` flags and copy the value of the flag if set into the rest config 🥳

pkg/test/assert.go
```diff
@@ -104,11 +104,16 @@ func Errors(namespace string, timeout int, errorFiles ...string) error {
 	return errors.New("error asserts not valid")
 }
 
-func Client(_ bool) (client.Client, error) {
+func Client(_ bool, as string) (client.Client, error) {
 	cfg, err := config.GetConfig()
```
porridge, line 108: How about:
- extract this and the next block (and an identical piece in `DiscoveryClient` below) into a new function in a new package
- in that package create an exported package-level variable `As` (or perhaps a bit more meaningful name) that defaults to an empty string.
- in that extracted code, use this `As` variable to tweak the config
- in the root cobra command, add the flag that sets the value of this new variable (I hope it's possible 🤔 )

porridge (commented):
I'm afraid plumbing this and related flags all the way this way is not going to be very sustainable 😟 ).

Can we try keeping the `Client` signature unchanged? I'm AFK so cannot easily check everything, but see an idea in an inline comment.

pkg/test/assert.go
logan-bobo: @porridge the method you have described is possible and much cleaner if you are happy with the appraoch (see updated) code I will roll this out to the other commands that generate a client config.

porridge: Sorry, perhaps I should have been more precise - when saying "this and the next block" above I meant just the following piece, not the whole `GetClient ` function:

```
	cfg, err := config.GetConfig()
	if err != nil {
		return nil, err
	}

	if ImpersonateAs != "" {
		cfg.Impersonate.UserName = ImpersonateAs
		}
```

As for the name of the package, perhaps something like `pkg/k8s` such that when imported as `client` the call would look like `k8s. Config()`? 🤔  WDYT?

porridge: But this looks very promising, thank you for wishing on this!

logan-bobo: Yeah I agree, we can also nest other k8s utilities under this package. I have moved the config generation to a function in this package.

pkg/kuttlctl/cmd/root.go
```diff
@@ -30,6 +32,7 @@ and serves as an API aggregation layer.
 		Version: version.Get().GitVersion,
 	}
 
+	cmd.PersistentFlags().StringVar(&k8s.ImpersonateAs, "as", "", "the username that you wish to impersonate")
```
logan-bobo, line 35: note as this is a `PersistentFlags` all dowstream commands will have access to this flag.

porridge, line 35: Let's mimic `kubectl`'s help strings.
```suggestion
	cmd.PersistentFlags().StringVar(&k8s.ImpersonateAs, "as", "", "Username to impersonate for the operation. User could be a regular user or a service account in a namespace.")
```

porridge (commented):
Excellent, this looks really neat!

Can you please:
- also change the call in `Harness.Config`
- make sure you update `docs/`
- since this is not very testable without specific cluster configuration, please copy-paste into PR description some terminal output showing how this fails without the option and works with it

docs/cli.md
```diff
@@ -61,6 +61,10 @@ Flags are:
 
   Directory to output kind logs to (if not specified, the current working directory).
 
+* **`--as`**
```
porridge, line 64:
```suggestion
* **`--as (string)`**
```

pkg/test/assert.go
```diff
@@ -105,27 +105,35 @@ func Errors(namespace string, timeout int, errorFiles ...string) error {
 }
 
 func Client(_ bool) (client.Client, error) {
-	cfg, err := config.GetConfig()
+	cfg, err := k8s.GetConfig()
+
```
porridge, line 109:
```suggestion
```

pkg/test/assert.go
```diff
@@ -105,27 +105,35 @@ func Errors(namespace string, timeout int, errorFiles ...string) error {
...
 	if err != nil {
 		return nil, err
 	}
+
 	client, err := testutils.NewRetryClient(cfg, client.Options{
 		Scheme: testutils.Scheme(),
 	})
+
```
porridge, line 117:
```suggestion
```

pkg/test/assert.go
```diff
@@ -105,27 +105,35 @@ func Errors(namespace string, timeout int, errorFiles ...string) error {
...
 		return nil, fmt.Errorf("fatal error getting client: %v", err)
 	}
+
 	return client, nil
 }
 
 func DiscoveryClient() (discovery.DiscoveryInterface, error) {
-	cfg, err := config.GetConfig()
+	cfg, err := k8s.GetConfig()
+
```
porridge, line 127:
```suggestion
```

pkg/test/assert.go
```diff
@@ -105,27 +105,35 @@ func Errors(namespace string, timeout int, errorFiles ...string) error {
...
 	if err != nil {
 		return nil, err
 	}
+
 	dclient, err := discovery.NewDiscoveryClientForConfig(cfg)
+
```
porridge, line 133:
```suggestion
```

pkg/k8s/main.go
```diff
@@ -0,0 +1,22 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+var ImpersonateAs = ""
+
+func GetConfig() (*rest.Config, error) {
+	cfg, err := config.GetConfig()
+
+	if err != nil {
+		return &rest.Config{}, err
```
porridge, line 14:
```suggestion
		return nil, err
```

porridge (approved):
Excellent ❤️ 
Just a few more nitpicks..

porridge:
Ah, 💩 
I applied my nitpicks to save some back-and-forth and forgot that this will break the DCO check 🤦🏻 
And now I cannot fix it easily because I cannot push to your fork. Please help! 😅

logan-bobo:
@porridge I will apply these and sign them

logan-bobo:
@porridge all of your comments have been added to the latest commit 🤝

PR #566, opened 2024-09-18
fix: handle redundant deletion of Namespace

### LINKED ISSUE #565: [Bug]: Deleting test namespace throws an error in-case of multiple TestSteps with lazy-loading Kubeconfigs
**What happened**:
While writing a test case with multiple `TestSteps`, we discovered that because in case of lazy-loading, we try to create the test namespace multiple times and subsequently schedule a deletion multiple times. See:
1. https://github.com/kudobuilder/kuttl/blob/c8c8342a15260011ce65b6f659c6adc67cb4a238/pkg/test/case.go#L382
2. https://github.com/kudobuilder/kuttl/blob/c8c8342a15260011ce65b6f659c6adc67cb4a238/pkg/test/case.go#L99-L129

The Namespace cleanup shouldn't throw an error.

Write two `TestStep` which lazy-load a Kubeconfig for the same K8s cluster. The test case would fail because the Namespace cleanup in the second `TestStep` would fail with a `NotFound` error.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.29
- KUTTL version (use `kubectl kuttl version`): 1.18
- OS (e.g. from /etc/os-release): MacOS
- Kernel (e.g. `uname -a`): Darwin
- Install tools:
- Others:

kumar-mallikarjuna:
Adds an `IsNotFound` check before attempting to delete the `Case`'s Namespace.

Fixes #565

### DIFF
--- a/pkg/test/case.go
+++ b/pkg/test/case.go
@@ -81,7 +81,9 @@ func (t *Case) DeleteNamespace(cl client.Client, ns *namespace) error {
 		},
 	}
 
-	if err := cl.Delete(ctx, nsObj); err != nil {
+	if err := cl.Delete(ctx, nsObj); k8serrors.IsNotFound(err) {
+		t.Logger.Logf("Namespace already cleaned up.")
+	} else if err != nil {
 		return err
 	}
 


pkg/test/case.go
```diff
@@ -81,7 +81,9 @@ func (t *Case) DeleteNamespace(cl client.Client, ns *namespace) error {
 		},
 	}
 
-	if err := cl.Delete(ctx, nsObj); err != nil {
+	if err := cl.Delete(ctx, nsObj); k8serrors.IsNotFound(err) {
+		t.Logger.Logf("Namespace already cleaned up")
```
porridge, line 85:
```suggestion
		t.Logger.Logf("Namespace already cleaned up.")
```

porridge (approved):

porridge:
Sorry, I made a mess with the sign-off again 🤦🏻

kumar-mallikarjuna:
Hey @porridge , might need to fix your sign-off here 😅

porridge:
> Hey @porridge , might need to fix your sign-off here 😅

Yes, but unfortunately I'm not able to do it myself. GitHub allows me to push a commit to your repo's branch based on a review suggestion (which is what I did, but used the wrong email) but it will not allow me to rewrite that commit. Sorry but I'll have to ask you to squash that commit into yours.

kumar-mallikarjuna:
Done, @porridge .

PR #576, opened 2024-11-23
[KEP-0009] feat: add expression based assertions

### LINKED ISSUE #562: Add CEL expression support
A Common Expression Language (CEL) based assertion semantic.

> The Common Expression Language (CEL) provides a standardized way to evaluate expressions, facilitating easier interoperability between different applications.

Reference: https://github.com/google/cel-spec

**Why is this needed**:

Currently, Kuttl lacks support for complex data manipulation in assertions. Here are a few examples where this limitation is apparent:
- Testing whether a Deployment has created _more than `n` Pods_.
  - While Kuttl supports equality testing, conditional matching poses a challenge.
- Slice assertion.
  - Kuttl's current method for handling slice comparison is somewhat obscure. Which becomes important especially when the order of elements in the slice is not fixed, which is common in Kubernetes (see https://github.com/kudobuilder/kuttl/issues/76#issuecomment-660944596).
  - Additionally, partially asserting slices is not possible (ref https://github.com/kudobuilder/kuttl/issues/76).

These issues could be resolved by incorporating a CEL engine into Kuttl, allowing for expression evaluation and more flexible assertions.

**How would this look like syntactically**:
```yaml
apiVersion: kuttl.dev/v1beta1
kind: TestAssert
celAssert:
  resources:
    - apiVersion: apps/v1
       kind: Deployment
       name: coredns
       namespace: kube-system
       id: resource
  expression: "resource.status.readyReplicas > 1"
```

---
Edit: Incorporate the comments from @porridge .

kumar-mallikarjuna:
This PR adds CEL-expression based assertions to `TestAsserts`. See https://github.com/kudobuilder/kuttl/blob/main/keps/0009-expression-based-assertions.md for more details.

Fixes #562


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -222,6 +226,19 @@ type TestCollector struct {
 	Cmd string `json:"command,omitempty"`
 }
 
+type TestResourceRef struct {
+	ApiVersion string `json:"apiVersion,omitempty"`
+	Kind       string `json:"kind,omitempty"`
+	Namespace  string `json:"namespace,omitempty"`
+	Name       string `json:"name,omitempty"`
+	Id         string `json:"id,omitempty"`
```
porridge, line 234: Should we call this `ref` like in the KEP?

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -152,6 +152,10 @@ type TestAssert struct {
 	Collectors []*TestCollector `json:"collectors,omitempty"`
 	// Commands is a set of commands to be run as assertions for the current step
 	Commands []TestAssertCommand `json:"commands,omitempty"`
+
+	ResourceRefs []TestResourceRef `json:"resourceRefs,omitempty"`
+
+	AssertExpressions AnyAllExpressions `json:"assertExpressions,omitempty"`
```
porridge, line 158: This does not follow the structure in the KEP: `assert(Any|All).celExpr`. Is it simply the prototype of an old iteration of the KEP?

pkg/test/utils/kubernetes.go
```diff
@@ -1321,3 +1398,29 @@ func Kubeconfig(cfg *rest.Config, w io.Writer) error {
...
+		Kind:    kind,
+	}
+	if len(apiVersion) > 1 {
+		gvk.Group = apiVersionSplit[0]
+	}
+
+	return gvk
+}
+
+func NewClient(kubeconfig, context string) func(bool) (client.Client, error) {
```
porridge, line 1415: This file is already very long as is. Can we perhaps move this (maybe along with `NewRetryClient`) into the `.../pkg/k8s` package?

pkg/test/utils/kubernetes.go
```diff
@@ -1321,3 +1398,29 @@ func Kubeconfig(cfg *rest.Config, w io.Writer) error {
 		},
 	}, w)
 }
+
+func constructGVK(apiVersion, kind string) schema.GroupVersionKind {
```
porridge, line 1402: This is a nice helper method, but it would be even more powerful it you
- changed it from a bare `func` to a method of `TestResourceRef`
- made it also internally create an `*Unstructured`:
  ```
		gvk := constructGVK(resourceRef.ApiVersion, resourceRef.Kind)
		referencedResource := &unstructured.Unstructured{}
		referencedResource.SetGroupVersionKind(gvk)
  ```
- and called it `NewUnstructured` or similar.

Thinking about it, it might make sense to have it also return a `NamespacedName` and an `error` in case crucial fields (like name and kind) are missing. On one hand, if we don't do any validation here, then the client would probably catch such garbage later. But on the other hand, we could use such validation function earlier, just after loading the assert files, in order to give user feedback on critical errors ASAP.

When you're at it, please also add a couple of test cases for it.

pkg/test/utils/kubernetes.go
```diff
@@ -1217,6 +1219,81 @@ func RunAssertCommands(ctx context.Context, logger Logger, namespace string, com
...
+		if err != nil {
+			return []error{fmt.Errorf("failed to evaluate program: %w", err)}
+		}
+
+		logger.Logf("expression '%v' evaluated to '%v'", expr, out.Value())
+		if out.Value() != true {
+			errs = append(errs, fmt.Errorf("failed validation, expression '%v' evaluated to '%v'", expr, out.Value()))
+		}
+	}
+
```
porridge, line 1293: Please make this a separate function that can be unit-tested without having to provide a k8s client.

porridge (changes requested):
Nice, but needs a few changes.

pkg/test/utils/kubernetes.go
```diff
@@ -1217,6 +1219,81 @@ func RunAssertCommands(ctx context.Context, logger Logger, namespace string, com
...
+	for _, expr := range expressions.Any {
+		ast, issues := env.Compile(expr)
+		if issues != nil && issues.Err() != nil {
+			return []error{fmt.Errorf("type-check error: %s", issues.Err())}
+		}
+
+		prg, err := env.Program(ast)
```
eddycharly, line 1278: It would be great to compile the program at load time, there's no point running the test if the compilation fails.

pkg/apis/testharness/v1beta1/test_types.go
kumar-mallikarjuna: In hind-sight,
```yaml
assert:
  any:
    ...
  all:
    ...
```
would be less verbose. WDYT?

pkg/test/utils/kubernetes.go
kumar-mallikarjuna: Good point but that creates an import cycle. :(

```shell
package command-line-arguments
        imports github.com/kudobuilder/kuttl/pkg/kuttlctl/cmd
        imports github.com/kudobuilder/kuttl/pkg/k8s
        imports github.com/kudobuilder/kuttl/pkg/test/utils
        imports github.com/kudobuilder/kuttl/pkg/k8s: import cycle not allowed
```

kumar-mallikarjuna: Maybe we could raise a separate PR to refactor the file?

pkg/apis/testharness/v1beta1/test_types.go
porridge: I think the verbosity is the nearly same - the number of letters is the same and the KEP version has one colon fewer.
And the KEP version is flatter, less indentation to think about, so I'd be for sticking to the KEP version. 🤷🏻

pkg/test/utils/kubernetes.go
porridge: Good idea. Without looking at details, I think `pkg/k8s` should be the leaf 🤔

pkg/test/utils/kubernetes.go
kumar-mallikarjuna: `NamespacedName`/`*unstructured.Unstructured` would be required when we evaluate the expressions. If the goal is also to validate, we should have two separate methods:
```go
func (t *TestResourceRef) BuildResourceReference() (namespacedName types.NamespacedName, referencedResource *unstructured.Unstructured)

func (t *TestResourceRef) Validate() (error)
```

pkg/test/utils/kubernetes.go
kumar-mallikarjuna: > I think pkg/k8s should be the leaf?

Yes, that makes sense.

pkg/test/step.go
```diff
@@ -533,6 +551,42 @@ func (s *Step) LoadYAML(file string) error {
...
+				if prg, err := assertion.BuildProgram(env); err != nil {
+					errs = append(errs, err)
+				} else {
+					s.Programs[assertion.CELExpression] = prg
+				}
+			}
+
+			if len(errs) > 0 {
+				return fmt.Errorf("failed to load TestAssert object from %s: %w", file, errors.Join(errs...))
+			}
```
porridge, line 589: Please extract this code into a `func loadPrograms(*harness.TestAssert) (map[string]cel.Program, error)` or similar and call it here.

pkg/test/utils/kubernetes.go
porridge: Now that this function is again only called from `(t *Case) Run()` it can go back to being an unexported `newClient` in `pkg/test/case.go`, right?

pkg/test/utils/kubernetes.go
```diff
@@ -1181,7 +1183,7 @@ func RunCommand(ctx context.Context, namespace string, cmd harness.Command, cwd
 	return nil, nil
 }
 
-func kubeconfigPath(actualDir, override string) string {
+func KubeconfigPath(actualDir, override string) string {
```
porridge, line 1186: This can remain unexported, right?

pkg/test/utils/expression.go
porridge, line 1: Perhaps this can go to a new package, e.g. `pkg/cel`? A `utils` package is an [anti-pattern](https://google.github.io/styleguide/go/best-practices#util-packages) so we shouldn't be making it worse 😅

pkg/test/utils/kubernetes.go
```diff
@@ -1217,6 +1219,76 @@ func RunAssertCommands(ctx context.Context, logger Logger, namespace string, com
...
+	assertAll []*harness.Assertion,
+) []error {
+	errs := []error{}
+	if len(assertAny) == 0 && len(assertAll) == 0 {
+		return errs
+	}
+
+	variables := make(map[string]interface{})
+	for _, resourceRef := range resourceRefs {
+		namespacedName, referencedResource := resourceRef.BuildResourceReference()
```
porridge, line 1238: Thinking more about the namespace, this should probably default to the current test's namespace if not specified in the assert file 🤔

pkg/test/utils/kubernetes.go
```diff
@@ -1217,6 +1219,76 @@ func RunAssertCommands(ctx context.Context, logger Logger, namespace string, com
 	return RunCommands(ctx, logger, namespace, convertAssertCommand(commands, timeout), workdir, timeout, kubeconfigOverride)
 }
 
+// RunAssertExpressions evaluates a set of CEL expressions specified as AnyAllExpressions
```
porridge, line 1222: Let's not make this file longer than it needs to be:
- please keep the part that populates `variables` inside `(s *Step) CheckAssertExpressions()`
- the rest (that does the evaluation and result checking) should probably go into a new cel package; please also add some unit tests for this part

porridge (changes requested):

pkg/apis/testharness/v1beta1/expression.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func (t *TestResourceRef) BuildResourceReference() (namespacedName types.NamespacedName, referencedResource *unstructured.Unstructured) {
+	referencedResource = &unstructured.Unstructured{}
+	apiVersionSplit := strings.Split(t.APIVersion, "/")
+	gvk := schema.GroupVersionKind{
+		Version: apiVersionSplit[len(apiVersionSplit)-1],
+		Kind:    t.Kind,
+	}
+	if len(t.APIVersion) > 1 {
```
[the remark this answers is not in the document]
kumar-mallikarjuna: 🤦  Thanks!

pkg/apis/testharness/v1beta1/expression.go
```diff
@@ -0,0 +1,75 @@
...
+		"apiVersion=%v, kind=%v, namespace=%v, name=%v, ref=%v",
+		t.APIVersion,
+		t.Kind,
+		t.Namespace,
+		t.Name,
+		t.Ref,
+	)
+}
+
+func (t *Assertion) BuildProgram(env *cel.Env) (cel.Program, error) {
```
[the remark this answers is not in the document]
kumar-mallikarjuna: Moved this to the new package `pkg/expressions`.

pkg/test/utils/expression.go
kumar-mallikarjuna: Done, thank you for the reference. I've named the package `expressions` since `cel` would conflict `cel-go`.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,135 @@
...
+	assertions = append(assertions, testAssert.AssertAny...)
+	assertions = append(assertions, testAssert.AssertAll...)
+
+	env, err := buildEnv(testAssert.ResourceRefs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build environment: %w", err)
+	}
+
+	var programs map[string]cel.Program
+	if len(assertions) > 0 {
```
porridge, line 118: Just return here if there are no assertions.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,135 @@
...
+	var errs []error
+	for _, resourceRef := range testAssert.ResourceRefs {
+		if err := resourceRef.Validate(); err != nil {
+			errs = append(errs, fmt.Errorf("validation failed for reference '%v': %w", resourceRef.String(), err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("failed to load resource reference(s): %w", errors.Join(errs...))
+	}
```
porridge, line 106: First validating ResourceRefs, then collecting assertions, and then going back to ResourceRefs to process them seems fragmented.
Can you please move these lines into `buildEnv` so that this is more streamlined?

pkg/expressions/cel.go
```diff
@@ -0,0 +1,135 @@
...
+
+	var programs map[string]cel.Program
+	if len(assertions) > 0 {
+		programs = make(map[string]cel.Program)
+	}
+
+	for _, assertion := range assertions {
+		if prg, err := buildProgram(assertion.CELExpression, env); err != nil {
+			errs = append(errs, err)
```
porridge, line 124: We should mention here which assertion expression in the file was problematic.

pkg/test/step.go
```diff
@@ -533,6 +566,11 @@ func (s *Step) LoadYAML(file string) error {
 			} else {
 				return fmt.Errorf("failed to load TestAssert object from %s: it contains an object of type %T", file, obj)
 			}
+
+			s.Programs, err = expressions.LoadPrograms(s.Assert)
+			if err != nil {
+				return fmt.Errorf("failed to load programs: %w", err)
```
porridge, line 572: The term "programs" are not mentioned in the API. We need to formulate error messages in terms that appear in the API. However this function processes both expressions and resourceRefs, so and I cannot think of a common term to describe them. If you cannot either, then perhaps this should propagate the error without wrapping, and `LoadPrograms` should take care of formulating a user-friendly error message... 🤔

pkg/test/step.go
```diff
@@ -422,6 +454,7 @@ func (s *Step) Check(namespace string, timeout int) []error {
 
 	if s.Assert != nil {
 		testErrors = append(testErrors, s.CheckAssertCommands(context.TODO(), namespace, s.Assert.Commands, timeout)...)
+		testErrors = append(testErrors, s.CheckAssertExpressions(context.TODO(), s.Assert.ResourceRefs, s.Assert.AssertAny, s.Assert.AssertAll)...)
```
porridge, line 457: I don't know why `Step.CheckAssertCommands()` takes `Step.Assert.Commands` as arguments instead of just accessing it directly in the body, but let's not copy this pattern here.

However you will likely need the `namespace` argument here, otherwise `CheckAssertExpressions` will not know which ephemeral namespace to retrieve resources from!

pkg/apis/testharness/v1beta1/expression.go
```diff
@@ -0,0 +1,65 @@
...
+		Name:      t.Name,
+	}
+
+	return
+}
+
+func (t *TestResourceRef) Validate() error {
+	apiVersionSplit := strings.Split(t.APIVersion, "/")
+	switch {
+	case t.APIVersion == "" || (len(apiVersionSplit) != 1 && len(apiVersionSplit) != 2):
```
porridge, line 43: This is simpler, and should be equivalent as `Split()` returns the input unchanged if the delimiter is not present, right?
```suggestion
	case t.APIVersion == "" || len(apiVersionSplit) > 2:
```

pkg/expressions/cel.go
```diff
@@ -0,0 +1,135 @@
...
+		prg, ok := programs[expr.CELExpression]
+		if !ok {
+			return []error{fmt.Errorf("couldn't find pre-built program for expression: %v", expr.CELExpression)}
+		}
+		out, _, err := prg.Eval(variables)
+		if err != nil {
+			return []error{fmt.Errorf("failed to evaluate program: %w", err)}
+		}
+
+		if out.Value() != true {
```
porridge, line 65: This might be a common part that is possible to extract as a function 🤔

pkg/expressions/cel.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to add resource parameter '%v' to environment: %w", resourceRef.Ref, err)
+		}
+	}
+
+	return env, nil
+}
+
+// RunAssertExpressions evaluates a set of CEL expressions specified as AnyAllExpressions
+func RunAssertExpressions(
```
porridge, line 43: There is little common code between the `all` and `any` case, let's split this into two separate functions.

pkg/test/expression_integration_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{
+			name:         "check multiple assert any with all failing",
+			runFailed:    true,
+			errorMessage: "no expression evaluated to true",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			step := buildTestStep(tc.name)
```
porridge, line 94: Please make `buildTestStep` a regular function rather than a closure, and pass the most granular `*T` to it.

```suggestion
			step := buildTestStep(t)
```

porridge (changes requested):
Thanks, this looks much better, but some more changes are needed. 🙏🏻 
Please see inline.

pkg/test/step.go
kumar-mallikarjuna: > However you will likely need the namespace argument here, otherwise CheckAssertExpressions will not know which ephemeral namespace to retrieve resources from!

`ResourceRefs` allow specifying the namespace right, so we can't be restricted to just the ephemeral `namespace`. WDYT?

porridge: The namespace in ResourceRefs should probably take precedence such that it's possible to assert things in other namespaces (statically named). But the user cannot foresee what ephemeral namespace name kuttl will come up with, so for assertions on objects in the current test's namespace (ResourceRef would not specify one in that case), we need to propagate it here, similar to how it's fed to `CheckAssertCommands`.

kumar-mallikarjuna: This is now incorporated in the latest commits. Please take a look when you have time. Thanks!

pkg/test/expression_integration_test.go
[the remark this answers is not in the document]
kumar-mallikarjuna: Done! :)

porridge (changes requested):
Almost there!
Just a bunch of nitpicks for messages and identifier names, and one issue in the test..

pkg/test/expression_integration_test.go
```diff
@@ -0,0 +1,172 @@
...
+	testutils "github.com/kudobuilder/kuttl/pkg/test/utils"
+)
+
+func buildTestStep(t *testing.T) *Step {
+	return &Step{
+		Name:   t.Name(),
+		Index:  0,
+		Logger: testutils.NewTestLogger(t, t.Name()),
+		Client: func(bool) (client.Client, error) {
+			return testenv.Client, nil
```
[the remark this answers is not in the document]
kumar-mallikarjuna: I couldn't declare one because there's already a `testenv` and `TestMain()` in the package 😓

pkg/test/step_integration_test_data/assert_expressions/check_expression_for_ephemeral_namespace/01-assert.yaml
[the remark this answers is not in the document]
kumar-mallikarjuna: I'm expecting the lexicographically last file to be the assert: 😅 

https://github.com/kudobuilder/kuttl/pull/576/files/a7122a69cc607624a9fb9186d04ab9a4cb11f02f#diff-f3050083c42f78c1718a3ff8105f31d3bac379304e4d8193438bedcae5a4f9e7R153

pkg/test/expression_integration_test.go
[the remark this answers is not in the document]
porridge: Maybe you could do something [like this](https://github.com/kudobuilder/kuttl/blob/65120e7038bae4d8896169dac0daa328a06fed89/pkg/test/case_integration_test.go#L20)?
I think we should change `step_integration_test.go` not to use a global, but this is outside the scope of this PR.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+		env, err = env.Extend(cel.Variable(resourceRef.Ref, cel.DynType))
+		if err != nil {
+			return nil, fmt.Errorf("failed to add resource parameter '%v' to environment: %w", resourceRef.Ref, err)
+		}
+	}
+
+	return env, nil
+}
+
+// RunAssertExpressions evaluates a set of CEL expressions specified as AnyAllExpressions
```
porridge, line 53:
```suggestion
// RunAssertExpressions evaluates a set of CEL expressions.
```

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+	return programs, nil
+}
+
+func evaluateExpression(expr string,
+	programs map[string]cel.Program,
+	variables map[string]interface{},
+) error {
+	prg, ok := programs[expr]
+	if !ok {
+		return fmt.Errorf("couldn't find pre-built parsed CEL expression '%v'", expr)
```
porridge, line 129:
```suggestion
		return fmt.Errorf("couldn't find pre-built parsed CEL expression %q", expr)
```

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	out, _, err := prg.Eval(variables)
+	if err != nil {
+		return fmt.Errorf("failed to evaluate CEL expression: %w", err)
+	}
+
+	if out.Value() != true {
+		return fmt.Errorf("expression '%v' evaluated to '%v'", expr, out.Value())
```
porridge, line 137:
```suggestion
		return fmt.Errorf("expression %q evaluated to %q", expr, out.Value())
```

pkg/test/expression_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+				"step_integration_test_data/assert_expressions/%s",
+				strings.ReplaceAll(tc.name, " ", "_"),
+			)
+
+			files, err := os.ReadDir(dirName)
+			assert.NoError(t, err)
+
+			for i := 0; i < len(files)-1; i++ {
+				fName := fmt.Sprintf("%s/%s", dirName, files[i].Name())
+				step := buildTestStep(t)
```
porridge, line 146: Only now I realized there is a separate `step` object for every file, and that this logic makes certain expectations about directory content. Also the directories look like kuttl tests but the test semantics are somewhat different.

Since the invalid expression case contains just one file, and all cases are effectively a single step, how about:
- using a single `step` object per test,
- with a bunch of `LoadYAML` calls, each using the same `expectLoadFailure` check
- and with a single `step.Run` call
- and drop the numeric prefix from filenames to avoid implying that there can be multiple steps.

pkg/test/expression_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+			Containers: []corev1.Container{
+				{
+					Name:  "metrics-server",
+					Image: "registry.k8s.io/metrics-server/metrics-server:v0.7.2",
+				},
+			},
+		},
+	}
+
+	assert.NoError(t, testenv.Client.Create(context.TODO(), codednsDeployment))
```
porridge, line 81: Here and below please use a `context.Background` or for bonus points, one with a sensible timeout.

pkg/test/step_integration_test_data/assert_expressions/check_expression_for_ephemeral_namespace/01-assert.yaml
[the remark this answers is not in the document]
porridge: Oh, that made me realize a few new things. See comments in the new review.

porridge (changes requested):
A few more things.

pkg/test/expression_integration_test.go
kumar-mallikarjuna: `check_expression_for_ephemeral_namespace` has two steps 🤔

kumar-mallikarjuna: Ah, I see. Incorporated now.

> and drop the numeric prefix from filenames to avoid implying that there can be multiple steps.

I've kept the prefixes for now in alignment with tests like `list-pods`. 😅

pkg/test/step.go
```diff
@@ -194,6 +194,7 @@ func (s *Step) Create(test *testing.T, namespace string) []error {
 
 	errors := []error{}
 
+	fmt.Println("HEREA1", len(s.Apply))
```
porridge, line 197: Leftovers? Please remove.

pkg/test/expression_integration_test.go
porridge: Note the ones used by `pkg/test/case_test.go` are OK to have prefixes, because they are ingested with `LoadTestSteps()` which is prefix-aware. Ideally we'd also do this here, and then assert `len(test.Steps) == 1` to prevent anyone who tries to add more steps from shooting themselves in the foot.

But I won't block on this, up to you.

porridge (changes requested):
LGTM now, just the debugging leftovers need to go :-)

pkg/test/step.go
kumar-mallikarjuna: My bad!

pkg/test/expression_integration_test.go
kumar-mallikarjuna: Okay, I see. Removed the prefixes. I also tried using `LoadTestSteps()` here but it's not super straightforward because we need to construct a `Suite` and its `TestReport` and then assert there are no errors in the report.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+		if err := resourceRef.Validate(); err != nil {
+			errs = append(errs, fmt.Errorf("validation failed for reference '%v': %w", resourceRef.String(), err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("failed to load resource reference(s): %w", errors.Join(errs...))
+	}
+
+	env, err := cel.NewEnv()
```
eddycharly, line 38: Are you extending the env somewhere ? Looks like you will want to register a bunch of CEL libs to enable features.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create environment: %w", err)
+	}
+
+	for _, resourceRef := range resourceRefs {
+		env, err = env.Extend(cel.Variable(resourceRef.Ref, cel.DynType))
```
eddycharly, line 44: Would be nice to use a type provider instead of `DynType`.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("failed to load expression(s): %w", errors.Join(errs...))
+	}
+
+	return programs, nil
+}
+
+func evaluateExpression(expr string,
+	programs map[string]cel.Program,
```
eddycharly, line 124: Why create a map to lookup the program corresponding to an expression ? 🤔

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -157,6 +157,11 @@ type TestAssert struct {
 	Collectors []*TestCollector `json:"collectors,omitempty"`
 	// Commands is a set of commands to be run as assertions for the current step
 	Commands []TestAssertCommand `json:"commands,omitempty"`
+
+	ResourceRefs []TestResourceRef `json:"resourceRefs,omitempty"`
```
eddycharly, line 161: You should consider two `TestResourceRef` with the same id an error.

pkg/expressions/cel.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if len(assertAny) != 0 && len(anyExprErrors) == len(assertAny) {
+		errs = append(errs, fmt.Errorf("no expression evaluated to true: %w", errors.Join(anyExprErrors...)))
+	}
+
+	if len(allExprErrors) > 0 {
+		errs = append(errs, fmt.Errorf("not all assertAll expressions evaluated to true: %w", errors.Join(allExprErrors...)))
+	}
+
+	return errs
```
eddycharly, line 86: This would make sense to return early here instead of evaluating all expressions before making a decision.

pkg/test/step.go
```diff
@@ -412,6 +416,28 @@ func (s *Step) CheckAssertCommands(ctx context.Context, namespace string, comman
...
+		if resourceRef.Namespace == "" {
+			resourceRef.Namespace = namespace
+		}
+		namespacedName, referencedResource := resourceRef.BuildResourceReference()
+		if err := client.Get(context.TODO(), namespacedName, referencedResource); err != nil {
+			return []error{fmt.Errorf("failed to get referenced resource '%v': %w", namespacedName, err)}
+		}
+
+		variables[resourceRef.Ref] = referencedResource.Object
+	}
```
eddycharly, line 436: This should be lazy loaded.

pkg/expressions/cel.go
porridge: Any suggestions on what libs are a must-have @eddycharly ?

pkg/expressions/cel.go
porridge: @eddycharly programs are compiled early at YAML loading time, so here we just access them. We use a map in order to mention the problematic expression in the error message in case it fails.

pkg/test/step.go
porridge: @eddycharly you mean as a performance improvement? Or do you see a correctness issue?

pkg/expressions/cel.go
eddycharly: I don't get it, if programs are compiled at load time you shouldn't even reach `evaluateExpression` right ?
I didn't look at it deeply but i would expect the focus is on the expression path, not the expression itself.

pkg/test/step.go
eddycharly: Somehow both. If the first condition is met in `any` why bother evaluating other conditions ?
I can imagine cases where the first condition returns true but the second condition throws an error. Returning early makes the evaluation succeed.

eddycharly: (this may or may not be what you want though)

porridge (changes requested):
Sorry for the delay in reviewing.

I think @eddycharly made a bunch of great suggestions. I know too little about CEL to judge with certainty, but I don't think any of these are really blocking. Perhaps we can do them as followup improvements, as the code looks OK in general, and this review is already taking waaaay longer than I expected. And it would feel good to get something usable merged already even if it's not perfect from the start. Unless you feel like addressing at least some of them straight away.

However one thing I completely forgot to mention earlier is documentation. Please at least update the [TestAssert schema description](https://github.com/kudobuilder/kuttl/blob/main/docs/testing/reference.md#testassert) and throw in a really short one-pager about this feature somewhere under `docs/testing` (probably a new file and link to it where appropriate). We can iterate on improving it later.

pkg/expressions/cel.go
porridge: An expression can compile OK (and its good to do it ASAP to catch syntax errors, etc), but evaluation can fail later (depending on resources in the cluster). An error message about the evaluation should mention the expression itself, so the user knows which expression was the problematic one.

kumar-mallikarjuna:
> I think @eddycharly made a bunch of great suggestions. I know too little about CEL to judge with certainty, but I don't think any of these are really blocking. Perhaps we can do them as followup improvements, as the code looks OK in general, and this review is already taking waaaay longer than I expected. And it would feel good to get something usable merged already even if it's not perfect from the start. Unless you feel like addressing at least some of them straight away.

Thanks for the suggestions, @eddycharly ! I'd vote to merge this as well and include improvements in a separate PR.

> However one thing I completely forgot to mention earlier is documentation. Please at least update the [TestAssert schema description](https://github.com/kudobuilder/kuttl/blob/main/docs/testing/reference.md#testassert) and throw in a really short one-pager about this feature somewhere under `docs/testing` (probably a new file and link to it where appropriate). We can iterate on improving it later.

Let me add this tomorrow during my day.

Thanks again @porridge , @eddycharly for the thorough review! :)

docs/testing/reference.md
```diff
@@ -102,15 +102,26 @@ commands:
...
 - type: pod
   pod: nginx
+resourceRefs:
+  - apiVersion: apps/v1
+    kind: Deployment
+    namespace: kube-system
+    name: coredns
+    ref: coredns
+assertAll:
+  - celExpr: "coredns.spec.replicas >= 2"
```
porridge, line 112: Let's keep the indentation consistent, and use a ref different from the name just to make it easier to see what is what from the example:
```suggestion
- apiVersion: apps/v1
  kind: Deployment
  namespace: kube-system
  name: coredns
  ref: coredns_deployment
assertAll:
- celExpr: "coredns_deployment.spec.replicas >= 2"
```

docs/testing/reference.md
```diff
@@ -168,3 +179,23 @@ skipLogOutput | bool   | If set, the output from the command is *not* logged. Us
...
+
+## Resource References
+
+The `Resource References` objects are used by `TestAssert` for declaring identifiers for expression based assertions.
+
+Field         |   Type | Description
+--------------|--------|---------------------------------------------------------------------
+apiVersion    | string | apiVersion of the target resource.
+kind    | string | Kind of the target resource.
+namespace    | string | Namespace of the target resource. When not specified, defaults to the auto-generated namespace.
```
porridge, line 191: Technically the namespace can be set statically by the user using a flag, so let's use a more generic description:
```suggestion
namespace    | string | Namespace of the target resource. When not specified, defaults to the  namespace of the current test.
```

porridge (approved):
LGTM apart from the two nitpicks inline. Please apply and let's merge this 🥳

kumar-mallikarjuna:
> LGTM apart from the two nitpicks inline. Please apply and let's merge this 🥳

Done, @porridge !

porridge (approved):

Issue #575, opened 2024-11-19
Optional more granular reporting for JUnit reports


tosky:
**What would you like to be added**:

https://github.com/kudobuilder/kuttl/pull/509 changed the format of the JUnit reports to be more granular. A statement at the end of the Caveats section says 

> If needed, I could add a knob to make this high-granularity opt-in.in the end 

This request is about introducing such knob to allow people to use the previous, less granular reporting format.

**Why is this needed**:

This potential need was already highlighted in the commit which changed the format, as highlighted above. I'd like to add that in the project I contribute, OpenStack Kubernetes Operators, we are starting to collect the results of KUTTL tests in a consistent way for internal reporting. The way each operator structured its tests is to have to have a basic mapping of 1-1 between tests and features. We don't want nor need to mark each step for reporting. So instead of having to post-process the JUnitXML file, it would be nice if kuttl could produce also (with some flag) the report in the old format.

porridge:
@tosky would you be able to verify that the knob added in #581 works as expected, by building the binary manually from tip of `main`?
I did check that with test-level granularity outputs from an example suite looks the same as it used to before #509 , but it wouldn't hurt to have a confirmation.
If that works, I can cut a release next week.

tosky:
Thank you so much for working on it! I've just tested the kuttl CLI from main and I confirm it new behavior when you use the new switch (either through the config file or through the command line parameter).

PR #580, opened 2024-11-25
chore: pre-factor code related to junit reporting

porridge:

This is preparation for changes needed to solve #575 .

There are no significant changes to functionality. See the individual commits, which:
1. change the report ordering to be stable (name-based) - this makes the subsequent test simple. I was somewhat surprised that the top-level (suite) order can differ, even though it's not executed in parallel. I don't think this is due to concurrency, but I'm adding a mutex just in case.
2. add a dummy test suite and golden files to prevent unintended changes to the reports
3. introduce an abstraction for reporting, that in a future PR will internally be creating the report slightly differently, based on a configuration setting

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,0 +1,22 @@
+name: "End-to-end tests"
+
+on:
+  push:
+    branches:
+      - main
+      - releases/*
+  pull_request:
+    branches:
+      - main
+      - releases/*
+
+jobs:
+  e2e-tests:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/setup-go@v2.1.4
+      with:
+        go-version: 1.21
+    - uses: actions/checkout@v2.4.0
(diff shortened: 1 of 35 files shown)


kovayur (approved):

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  e2e-tests:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/setup-go@v2.1.4
+      with:
+        go-version: 1.21
+    - uses: actions/checkout@v2.4.0
+    - name: "Run end-to-end tests"
+      run: make e2e-test INTEGRATION_OUTPUT_JUNIT="true"
```
kovayur, line 22: Does `INTEGRATION_OUTPUT_JUNIT` have any effect?

porridge: Thanks for pointing this out, leftover from copy-paste 😅

PR #582, opened 2024-11-29
Bump go version to 1.22.7

avorima:

Increase the minimum go version to 1.22.7 which includes some CVE fixes.

For context: we are building images with the github release binary and our image scanner picks up the CVEs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/kudobuilder/kuttl
 
-go 1.22.0
-
-toolchain go1.22.1
+go 1.22.7
 
 require (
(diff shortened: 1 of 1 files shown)


porridge (approved):
Looks like it's time for me to finally read up on the semantics of these directives. It seems strange that having a stale version here would prevent someone from building with a _more recent_ toolchain 🤯

avorima:
Depends on how the binaries are built. Our workflows configure the setup-go action to read the exact version from the go.mod, but the version and toolchain are always a minimum requirement. I don't know how kuttl release binaries are built, so this PR might've been redundant.

PR #583, opened 2024-11-29
Clarify help terminology

gmolau:
Clarifies the output of `kuttl test --help` in two ways, see comments for details.

### DIFF
--- a/pkg/kuttlctl/cmd/test.go
+++ b/pkg/kuttlctl/cmd/test.go
@@ -67,7 +67,7 @@ func newTestCmd() *cobra.Command { //nolint:gocyclo
 	options := harness.TestSuite{}
 
 	testCmd := &cobra.Command{
-		Use:   "test [flags]... [test directories]...",
+		Use:   "test [flags]... [test suite]...",
 		Short: "Test KUTTL and Operators.",
 		Long: `Runs integration tests against a Kubernetes cluster.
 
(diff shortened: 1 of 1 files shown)


pkg/kuttlctl/cmd/test.go
```diff
@@ -67,7 +67,7 @@ func newTestCmd() *cobra.Command { //nolint:gocyclo
 	options := harness.TestSuite{}
 
 	testCmd := &cobra.Command{
-		Use:   "test [flags]... [test directories]...",
+		Use:   "test [flags]... [test suite]...",
```
gmolau, line 70: Directories in kuttl have different semantic meaning, they can be a suite or a case. This uses the correct terminology so make it clear that the suite is expected here, passing a case does not work.

pkg/kuttlctl/cmd/test.go
```diff
@@ -249,7 +249,7 @@ For more detailed documentation, visit: https://kuttl.dev`,
 	testCmd.Flags().StringVar(&configPath, "config", "", "Path to file to load base test settings from (these may be overridden with command-line arguments).")
 	testCmd.Flags().StringVar(&crdDir, "crd-dir", "", "Directory to load CustomResourceDefinitions from prior to running the tests.")
 	testCmd.Flags().StringSliceVar(&manifestDirs, "manifest-dir", []string{}, "One or more directories containing manifests to apply before running the tests.")
-	testCmd.Flags().StringVar(&testToRun, "test", "", "If set, the specific test case to run.")
+	testCmd.Flags().StringVar(&testToRun, "test", "", "If set, the specific test case to run (basename of the test case dir).")
```
gmolau, line 252: Clarifies that only the basename is expected here, not the relative or absolute path.

porridge (approved):
Thanks!
But please amend the commit to contain the DCO line (see failing check for instructions).

gmolau:
@porridge Thanks, done now

porridge:
The [e2e failure is unrelated](https://github.com/kudobuilder/kuttl/issues/584), merging.

PR #587, opened 2024-12-03
Allow for processing tls-server-name attribute in kubeconfig context

### LINKED ISSUE #586: Support for tls-server-name kubeconfig property
There exists a property `tls-server-name` available to a kubeconfig context ( [docs](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#Cluster) )
It is supported in the [kubernetes client library](https://github.com/kubernetes/client-go/blob/37045084c2aa82927b0e5ffc752861430fd7e4ab/plugin/pkg/client/auth/exec/exec_test.go#L701)

I would like support for this to be added for this flag

**Why is this needed**:
When using security software to control access to kube-apiservers kuttl cannot currently be used.

jwitko:
Fixes #586 by allowing for processing of the `tls-server-name` attribute from the current kubecontext.


porridge (approved):
Thank you!

PR #590, opened 2024-12-04
refactor: moves kubernetes utilities to `pkg/kubernetes`

### LINKED ISSUE #589: Refactor `pkg/test/utils/kubernetes.go`
A new package `pkg/kubernetes/`.

**Why is this needed**:
`pkg/test/utils/kubernetes.go` has grown huge and we should refactor it to a separate package.

Reference discussion: https://github.com/kudobuilder/kuttl/pull/576#discussion_r1859982831

cc: @porridge

kumar-mallikarjuna:
- Renames pkg/k8s to pkg/kubernetes
- Moves the utilities in `pkg/test/utils/kubernetes.go` to `pkg/kubernetes/`.

```shell
$ tree pkg/kubernetes 
pkg/kubernetes
├── kubernetes.go
├── retry_client.go
├── retry_client_test.go
├── scheme.go
├── serialization.go
├── setup.go
├── unstructured.go
└── wait.go

1 directory, 8 files
```

Fixes #589

cc: @porridge


pkg/test/utils/kubernetes_integration_test.go
```diff
@@ -17,14 +17,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	harness "github.com/kudobuilder/kuttl/pkg/apis/testharness/v1beta1"
+	"github.com/kudobuilder/kuttl/pkg/kubernetes"
 )
 
-var testenv TestEnvironment
+var testenv kubernetes.TestEnvironment
 
 func TestMain(m *testing.M) {
```
porridge, line 25: This test should probably be moved to `pkg/kubernetes`?

pkg/test/utils/kubernetes_integration_test.go
```diff
@@ -38,12 +39,12 @@ func TestCreateOrUpdate(t *testing.T) {
 	// Run the test a bunch of times to try to trigger a conflict and ensure that it handles conflicts properly.
```
porridge, line 39: Ditto

pkg/test/utils/kubernetes_integration_test.go
```diff
@@ -63,23 +64,23 @@ func TestCreateOrUpdate(t *testing.T) {
...
 
-		_, err = CreateOrUpdate(context.TODO(), testenv.Client, SetAnnotation(depToUpdate, "test", "hello"), true)
+		_, err = kubernetes.CreateOrUpdate(context.TODO(), testenv.Client, kubernetes.SetAnnotation(depToUpdate, "test", "hello"), true)
 		assert.Nil(t, err)
 
 		quit <- true
 	}
 }
 
 func TestClientWatch(t *testing.T) {
```
porridge, line 82: Ditto

pkg/test/utils/kubernetes_test.go
```diff
@@ -3,24 +3,20 @@ package utils
...
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	harness "github.com/kudobuilder/kuttl/pkg/apis/testharness/v1beta1"
+	"github.com/kudobuilder/kuttl/pkg/kubernetes"
 )
 
 func TestNamespaced(t *testing.T) {
```
porridge, line 18: Ditto.

pkg/test/utils/kubernetes_test.go
```diff
@@ -140,33 +87,6 @@ func TestKubeconfigPath(t *testing.T) {
...
-func TestRetryWithTimeout(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-	defer cancel()
-
-	assert.Equal(t, errors.New("error"), Retry(ctx, func(context.Context) error {
-		return errors.New("error")
-	}, func(err error) bool { return true }))
-}
-
 func TestLoadYAML(t *testing.T) {
```
porridge, line 90: Ditto.

pkg/test/utils/kubernetes_test.go
```diff
@@ -517,14 +437,14 @@ func TestRunScript(t *testing.T) {
 }
 
 func TestPrettyDiff(t *testing.T) {
```
porridge, line 439: Ditto.

pkg/test/utils/kubernetes.go
porridge, line 1: The `func Kubeconfig(cfg *rest.Config, w io.Writer) error {` func should probably also land in `pkg/kubernetes`.

kumar-mallikarjuna:
I've incorporate the suggested changes. Thanks for the quick review, @porridge !

kumar-mallikarjuna:
@porridge , I don't understand why the UTs are failing. 🤔 Any clues?

kumar-mallikarjuna:
> > @porridge , I don't understand why the UTs are failing. 🤔 Any clues?
> 
> ```
> 2024/12/05 17:01:06 unable to start control plane itself: failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
> ```
> 
> Perhaps you mixed up integration tests (which require envtest) with unit tests (which can run without it)? 🤔

The `//go:build integration` directive was missing. 😄 Can you please take a look again, @porridge ?

pkg/test/utils/kubernetes_integration_test.go
porridge, line 1: Looks like all remaining tests from this file can be moved to the file which is currently called `kubernetes_test.go` (but see the comment there) since none of them need envtest any more, right?
Then delete this file.

pkg/test/utils/kubernetes_test.go
```diff
@@ -140,109 +33,6 @@ func TestKubeconfigPath(t *testing.T) {
...
-						"image": "nginx:1.7.9",
-						"name":  "nginx",
-					},
-				},
-			},
-		},
-	}, objs[1])
-}
-
 func TestMatchesKind(t *testing.T) {
```
porridge, line 36: Looks like this test should also move to `pkg/kubernetes`?

pkg/test/utils/kubernetes_test.go
porridge, line 1: Looks like this can be renamed to `commands_test.go`.

pkg/kubernetes/setup.go
```diff
@@ -0,0 +1,134 @@
...
+	Client          Client
+	DiscoveryClient discovery.DiscoveryInterface
+}
+
+// Client is the controller-runtime Client interface with an added Watch method.
+type Client interface {
+	client.Client
+	// Watch watches a specific object and returns all events for it.
+	Watch(ctx context.Context, obj runtime.Object) (watch.Interface, error)
+}
```
porridge, line 134: I think this particular function an the two types could actually stay in `pkg/test/utils` but perhaps move to a file `environment.go` or `setup.go`? WDYT?

pkg/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	if serviceAccountName != "" {
+		pod.Spec.ServiceAccountName = serviceAccountName
+	}
+	return &pod
+}
+
+// FakeDiscoveryClient returns a fake discovery client that is populated with some types for use in
+// unit tests.
+func FakeDiscoveryClient() discovery.DiscoveryInterface {
```
porridge, line 139: This function seems to be used only by tests. Please move it to a separate package such as `pkg/kubernetes/fake`. It might then make sense to drop the `Fake` prefix from the function name to avoid stuttering.

pkg/kubernetes/kubernetes.go
porridge, line 1: Please move `ImpersonateAs`, `GetConfig`, `BuildConfigWithContext` and `Kubeconfig` to a new file called `config.go` or similar. Please also move associated tests from `kubernetes_test.go` to `config_test.go`.

Then (after also moving `FakeDiscoveryClient` as mentioned below) please rename this one to something like `objects.go` (and then also rename `kubernetes_test.go` accordingly.

porridge (changes requested):
Sorry it took me so long to review, the recent weeks were rather hectic. 😫 

I really appreciate you taking upon yourself to address this tech debt!

This is looking almost perfect, just a few more improvement requests inline, and I think this will be good to merge! ❤️

pkg/kubernetes/setup.go
kumar-mallikarjuna: > but perhaps move to a file environment.go or setup.go? WDYT?

These are in `setup.go` already o.O

porridge: I mean `pkg/test/utils/setup.go`.
The mention of `envtest` in `StartTestEnvironment` makes me think like `pkg/test/utils` might be a better home for these than `pkg/kubernetes`....

kumar-mallikarjuna: Ah okay, I see makes sense. Incorporated.

kumar-mallikarjuna: @porridge , yikes. That adds an import cycle because of the `RetryClient`. 💀 How about `pkg/kubernetes`? It's confusing to keep it in `pkg/test/utils` I think - because `pkg/test/utils` has the utils for "Kuttl test" semantics and not the "integration/unit tests." :D WDYT?

porridge: Just keep this in `pkg/kubernetes/setup.go`  if moving is problematic, it's not a big issue. I'll take a look after it's merged.

kumar-mallikarjuna: Done

porridge (approved):
Excellent! ❤️ ❤️ ❤️

PR #613, opened 2025-02-26
chore: refactoring for the Case type

porridge:

This PR may look intimidating, but 99% of it is just mechanical moves and renames. I highly recommend reviewing commit by commit.

Here is a high level overview:
- extracted event reporting logic and file collection logic into separate packages, to keep the file smaller and more focused. As a result some `test_data` files are copied, since they are also used by other tests in the original package.
- a fix to an error message
- unexported methods where possible
- rename the receiver for `*Case` to be `c` rather than the confusing `t`
- moved the event reporting suppression logic out of the `Run` method
- extracted the setup code from `Run` into a separate method to keep the latter more focused and shorter
- moved `new*Client` functions into the pkg/kubernetes package
- renamed some variables and fields in Case struct
- extracted some Step setup code into a Setup method in Step type

This prepares ground an upcoming [change](https://github.com/kudobuilder/kuttl/pull/387) (the part of which moves namespace determination to happen earlier - from `Run()` to `LoadTestSteps()`) along with some long overdue cleanups.


vladbologa (approved):
Maybe not sure about the duplication of the test data, but LGTM.

PR #620, opened 2025-03-25
feat: add as-groups as cli flag

oliverbaehler:

Adding impersonation via Flag, it's a bit easier than having to manage n-kubeconfigs are always rewriting the kubeconfig. My cases require to always change the impersonation. 

Fixes #


porridge (changes requested):
Actually, please also update `docs/cli.md` and this will be good to merge.

PR #632, opened 2025-10-14
chore: move packages to internal

porridge:

We should not be exposing all code as a public API.

Leave just the apis directory and the Harness type in pkg/ to avoid the single active user that I could find.

When doing this, flatten the structure by moving packages under the `pkg/test` package directly into `internal`.

### DIFF
--- a/docs/api-integration.md
+++ b/docs/api-integration.md
@@ -39,9 +39,7 @@ func TestWithKuttl(t *testing.T) {
       TestSuite: options,
       T:         t,
     }
-
-    harness.Setup()
-	harness.Run()
+    harness.Run()
 }
 
 ```


docs/api-integration.md
```diff
@@ -37,16 +34,15 @@ options := harness.TestSuite{}
...
     harness := test.Harness{
       TestSuite: options,
       T:         t,
     }
 
-    harness.Run()
-  })
-},
+    harness.Setup()
+	harness.Run()
```
vladbologa:
```suggestion
    harness.Setup()
    harness.RunTests()
```

`harness.Run()` calls `Setup()` and `RunTests()`.

porridge: Ah, good point, looks like I followed not the best example, calling Run is suffcient, then.

PR #634, opened 2025-10-19
chore: upgrade k8s lib and controller-runtime

davidebianchi:

Upgrade go version to 1.24, controller-runtime to 0.22 and kubernetes libs to 0.34

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -133,7 +133,8 @@ linters-settings:
         funcs:
           - (github.com/kudobuilder/kuttl/internal/utils.Logger).Logf
           - (github.com/kudobuilder/kuttl/internal/utils.TestLogger).Logf
-          - (github.com/kudobuilder/kuttl/internal/case.Case).logkcf
+          - (github.com/kudobuilder/kuttl/internal/case.clientWithKubeConfig).Logf
+          - (github.com/kudobuilder/kuttl/internal/case.clientWithKubeConfig).Wrapf
   # TODO: fix the `context` usage in step package and enable these linters.
   usetesting:
     context-background: false
--- a/docs/testing/reference.md
+++ b/docs/testing/reference.md
@@ -42,11 +42,14 @@ suppress          | list of strings  | Suppresses log collection of the specifie
 
 ## TestStep
 
-The `TestStep` object can be used to specify settings for a test step and can be specified in any test step YAML.
+The `TestStep` object can be used to specify settings for a test step and can be specified in any test step YAML
+file other than the assert and errors files.
 
 ```yaml
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
+metadata:
+  name: example-step
 apply:
 - my-new-resource.yaml
 assert:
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -12,7 +12,7 @@ BUILD_DATE := $(shell date -u -d "@$SOURCE_DATE_EPOCH" "+${DATE_FMT}" 2>/dev/nul
 LDFLAGS := -X ${GIT_VERSION_PATH}=${GIT_VERSION} -X ${GIT_COMMIT_PATH}=${GIT_COMMIT} -X ${BUILD_DATE_PATH}=${BUILD_DATE}
 
 # Please update the list of linters in .golagci.yml when bumping the version.
```
porridge: Anything interesting in the linter list?

porridge (commented):
Thanks for the bumps @davidebianchi !
I just merged a massive refactoring, and there are another couple of changes in flight that touch `go.mod` so perhaps it's best if I clone your PR (with correct attribution) after I'm done with them?
I certainly want to bump the deps before making a release sometime this or next week.

Makefile
davidebianchi: Rebased

.golangci.yml
```diff
@@ -131,6 +134,9 @@ linters-settings:
           - (github.com/kudobuilder/kuttl/internal/utils.Logger).Logf
           - (github.com/kudobuilder/kuttl/internal/utils.TestLogger).Logf
           - (github.com/kudobuilder/kuttl/internal/case.Case).logkcf
+  usetesting:
```
porridge:
```suggestion
  # TODO: fix the `context` usage in step package and enable these linters.
  usetesting:
```

porridge (changes requested):
This looks great! Thanks a lot for the cleanups! ❤️ 
Just two nitpicks inline. 🙏🏻

internal/kubernetes/serialization_test.go
```diff
@@ -139,13 +140,11 @@ func TestPrettyDiff(t *testing.T) {
 +  observedGeneration: 2
 +  replicas: 1
 +  unavailableReplicas: 1
-+  updatedReplicas: 1
- 
-`, result)
++  updatedReplicas: 1`, strings.TrimSpace(result))
```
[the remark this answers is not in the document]
davidebianchi: It is needed both a newline and a space, and the space in some ide is auto trimmed by default. But got it, I'll restore it

internal/utils/commands_test.go
```diff
@@ -276,18 +269,21 @@ func TestRunCommand(t *testing.T) {
...
+	cmd, err = RunCommand(t.Context(), "", hcmd, "", stdout, stderr, logger, 0, "")
 	assert.NoError(t, err)
 	assert.NotNil(t, cmd)
-	// no stdout for background processes
-	assert.Empty(t, strings.TrimSpace(stdout.String()))
+
+	// Wait for background process to complete before checking stdout
+	err = cmd.Wait()
+	assert.NoError(t, err)
+	assert.Equal(t, "hello", strings.TrimSpace(stdout.String()))
```
davidebianchi: Hi @porridge, it seems that there was a race condition in `TestRunCommand` test, since the test is trying to access to stdout result without wait the end of the process (so, some times the unit test pipeline could fail).
I think this change should be enough to test the background run.
What do you think about run unit tests with `-race` option in pipeline?

porridge (approved):

PR #635, opened 2025-10-20
chore: second round of pre-factoring

porridge:
 
This is the last batch (I promise!) of pre-factoring before the actual change which moves determination of namespace name from Run() to creation.

No user-affecting changes except for some log/error message improvements.

**Again, reviewing commit-by-commit is recommended.**

### DIFF
--- a/internal/testcase/case.go
+++ b/internal/testcase/case.go
@@ -265,6 +265,10 @@ func (c *Case) setup(test *testing.T) error {
 }
 
 // Returns clients for all steps other than the lazy loaded ones.
+// The returned map will always contain at least a default client with a key of empty string.
+// However, there may be more pairs, since each step may optionally specify a path to kubeconfig that should be used.
+// This is useful for multi-cluster or multi-context tests.
+// For those, the client will be keyed with the specified path to kubeconfig file.
 func (c *Case) getEagerClients() (map[string]client.Client, error) {
 	defaultClient, err := c.getClient(false)
 	if err != nil {


internal/testcase/case.go
```diff
@@ -109,27 +110,28 @@ func (c *Case) deleteNamespace(cl client.Client) error {
 	}
 
 	if err := cl.Delete(ctx, nsObj); k8serrors.IsNotFound(err) {
-		c.logger.Logf("Namespace already cleaned up.")
+		c.logger.Logf(maybeAppendKubeConfigInfo("Namespace %q already cleaned up.", kubeconfigPath), c.ns.name)
 	} else if err != nil {
-		return err
+		return errors.Wrapf(err, maybeAppendKubeConfigInfo("failed to delete namespace %q", kubeconfigPath), c.ns.name)
```
mclasmeier: It is my understanding that modern Go code should migrate to using `fmt.Errorf` instead of relying on `errors.Wrapf`. I am raising this since the current kuttl codebase doesn't seem to be using `Wrapf` a lot (just found three occurrences, seems all three are from this PR). Hence my suggestion to adopt `Errorf`.

mclasmeier (changes requested):
Review finished.
Overall looking good, some things about format strings seem to require fixing + I have some style suggestions.

internal/testcase/case.go
vladbologa: I agree, the codebase already uses `fmt.Errorf` for wrapping - but in some cases not ideally, e.g. `fmt.Errorf("fatal error getting client: %v", err)` with `%v` instead of `%w`.

porridge:
BTW, now I remember why I always preferred `errors.Wrapf`: it correctly returns `nil` when the wrapped error is `nil`... With `fmt.Errorf` one needs extra lines to cover that case 😟

PR #637, opened 2025-10-27
chore: deterimine namespace name early

porridge:

The primary goal is making the namespace name available at the time when `LoadTestSteps()` is called, so that it can be used as a test file templating parameter.

Note:
- the `Skipping creation of user-supplied namespace` message is removed since it's superfluous given the `Namespace %q already exists` message.
- namespace existence is established indirectly at namespace creation time, rather than using a separate explicit GET. This makes it a tiny bit faster, but more importantly this knowledge is now maintained correctly on a per-kubeconfig basis.
- at first sight, this change seems to remove the `Skipping deletion of user-supplied namespace...` message, but in reality that was dead code, since in the !autocreated case `deleteNamespace` was never scheduled to run. So instead we're adding a more specific `Skipping deletion of pre-existing user supplied namespace` message.
- scheduling namespace creation using test.Cleanup() is now done after, rather than before the CREATE call, so there is a minute chance it will not be cleaned up if, in the time window between the CREATE call and the Cleanup call, the test is aborted externally. I think the simplicity of the code is worth the risk of a leftover empty namespace, given that when a kuttl run is aborted, some garbage might be left anyway.

### DIFF
--- a/internal/testcase/case.go
+++ b/internal/testcase/case.go
@@ -159,33 +159,8 @@ func (c *Case) createNamespace(test *testing.T, cl client.Client, kubeconfigPath
 	})
 	if !c.skipDelete {
 		test.Cleanup(func() {
-			// The subtlety of whether a namespace should be cleaned up or not becomes obvious when we consider
-			// that each test step can use a different kubeconfig. There is no easy way to tell whether the clusters
-			// behind these kubeconfigs overlap, especially when using lazy kubeconfig loading. If they do, then there
-			// is no clear client <-> namespace ownership relation.
-			//
-			// Historically the rules that kuttl obeyed had been:
-			// - no user-specified namespace: make ephemeral namespace name, for every client: create it (no error if
-			//   already present), and clean it up afterward
-			// - otherwise:
-			//   - BEFORE the test, check if the supplied namespace exists, but USING THE DEFAULT CLIENT ONLY,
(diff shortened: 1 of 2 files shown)


internal/testcase/case.go
```diff
@@ -161,8 +157,47 @@ func (c *Case) createNamespace(test *testing.T, cl client.Client, kubeconfigPath
...
+			//   kubeconfigs.
+			// - the existence of namespace for the default client should not determine (at least not unconditionally)
+			//   the fate of the namespace for other clients.
+			// - in particular, if a namespace was missing on default cluster, but pre-existed on auxiliary cluster,
+			//   kuttl would clean the latter up. If it was the other way round, it would not touch any namespace,
+			//   and likely fail due to missing namespace on the auxiliary cluster.
+			//
+			// Here we try to be more thoughtful, by maintaining the presence/absence information separately for every
+			// client. However, for backward compatibility we still take into account whether the namespace was user
+			// supplied or not.
```
vladbologa: Could this be moved to `keps/0008-multi-cluster.yaml` (or a new one) and just reference it here?

And then this comment could be something like:
```
        // Namespace cleanup is tracked per-client for multi-cluster tests.
        // See KEP-0008 for details on backward compatibility decisions.
```

porridge: Great idea! Moved.

PR #638, opened 2025-11-03
chore: replace logkcf with Logf in a new type

porridge:

Addresses feedback from https://github.com/kudobuilder/kuttl/pull/635#discussion_r2464762092.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -130,8 +130,8 @@ linters-settings:
         funcs:
           - (github.com/kudobuilder/kuttl/internal/utils.Logger).Logf
           - (github.com/kudobuilder/kuttl/internal/utils.TestLogger).Logf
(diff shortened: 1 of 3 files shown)


porridge:
I also updated the linter config and made sure it works:

```
./bin/golangci-lint --timeout 5m run --build-tags integration
internal/testcase/case.go:120:10: printf: (github.com/kudobuilder/kuttl/internal/testcase.clientWithPath).Logf call has arguments but no formatting directives (govet)
		cl.Logf("Namespace q already cleaned up.", c.ns.name)
		       ^
internal/testcase/case.go:122:18: printf: (github.com/kudobuilder/kuttl/internal/testcase.clientWithPath).Wrapf call has arguments but no formatting directives (govet)
		return cl.Wrapf(err, "failed to delete namespace q", c.ns.name)
		               ^
make: *** [Makefile:42: lint] Error 1
```

PTAL @vladbologa

PR #682, opened 2026-04-23
fix: add random suffix to namespace name to prevent collisions

### LINKED ISSUE #500: Auto-generated namespace name clashes

mandre:
Append a 4-character random suffix to the generated namespace name and drop the `test-` infix to keep the max length unchanged.

    Old format: kuttl-test-<adjective>-<noun>
    New format: kuttl-<adjective>-<noun>-<4 char suffix>
    Example:    kuttl-enormous-termite-x7kg

This increases the namespace space from ~200k to ~40 billion combinations, making collisions statistically negligible even at high parallelism.

Fixes https://github.com/kudobuilder/kuttl/issues/500.

### DIFF
--- a/internal/testcase/case.go
+++ b/internal/testcase/case.go
@@ -15,6 +15,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -149,7 +150,7 @@ func NewCase(name string, parentPath string, options ...CaseOption) *Case {
 
 	if c.ns == nil {
 		c.ns = &namespace{
-			name:         fmt.Sprintf("kuttl-test-%s", petname.Generate(2, "-")),
+			name:         fmt.Sprintf("kuttl-%s-%s", petname.Generate(2, "-"), utilrand.String(4)),
 			userSupplied: false,
 		}
 	}


porridge (commented):
First of all, thank you for the contribution, I really appreciate the effort.

However to be honest, I'm really worried about the complexity this approach introduces. The logic that generates, creates and cleans up namespaces is already quite complicated, and trying to refactor it caused a [regression](https://github.com/kudobuilder/kuttl/issues/656) once. Adding a feedback loop in the middle is too much, I'm afraid.  Also, reactive retrying gets less and less effective the higher the number of parallel test cases is.

However, I can also very well relate to the annoyance that a name collision might cause. Could we try a different approach first to see whether it would be sufficient? For example replace this [petname call](https://github.com/kudobuilder/kuttl/blob/a57b836d05f6cf5c83b098857e085a44688cd219/internal/testcase/case.go#L152) with a call for a single petname noun (so that in a low parallelism case the names are still easily distinguishable), and instead of the adjective use `k8s.io/apimachinery/pkg/util/rand.String(8)` (so that we don't increase the possible max length)?

If my calculations are right it should increase the number of possible namespace names from about 200,000 (~450²) to over 1E14 (~450*8^27).

WDYT?

mandre:
> However, I can also very well relate to the annoyance that a name collision might cause. Could we try a different approach first to see whether it would be sufficient? For example replace this [petname call](https://github.com/kudobuilder/kuttl/blob/a57b836d05f6cf5c83b098857e085a44688cd219/internal/testcase/case.go#L152) with a call for a single petname noun (so that in a low parallelism case the names are still easily distinguishable), and instead of the adjective use `k8s.io/apimachinery/pkg/util/rand.String(8)` (so that we don't increase the possible max length)?

If we don't care about keeping the exact naming pattern (which my PR tried to do), then perhaps the simplest would be to add a small random suffix to decrease the chances of a conflict, while not completely eliminating them.
I personally find the current naming cute, with adjective + animal (reminds me of the default docker container names that have adjective + famous scientist) so my preference (if it counts :D) would be to keep that and simply add a suffix.

Are we worried about keeping the max length the same? If my math is correct, it's currently 28 chars, well below the max 63 chars limit for k8s namespaces.

porridge:
> Are we worried about keeping the max length the same? If my math is correct, it's currently 28 chars, well below the max 63 chars limit for k8s namespaces.

Yes, some operators use the namespace name as a prefix for some cluster-scoped resources to avoid clashing with other instances of itself installed in a different namespace.

I did some birthday paradox calculations for a few cases, the quick summary of the chance of a collision is:

| scheme | test cases: 151  | 1k  | 5k | 10k
| ----------- | ---------------------- | ------------------ | -------------- | ------------
| `kuttl-test-<adjective>-<noun>` (current) | 5% | 91% | ~100% | ~100%
| `kuttl-<adjective>-<noun>-<4 char suffix>` | 0.00001% | 0.00046% | 0.01161% | 0.04645%
| `kuttl-test-<noun>-<8 char suffix>` | 🎲 | 0.0000004% | 0.00001% | 0.00004%
| `kuttl-<noun>-<13 char suffix>` | 🎲 | 🎲 | 🎲 | 🎲

(🎲 meaning probability lower than winning the powerball jackpot: $$\approx 1/3 * 10^8$$)

So, definitely the current scheme is terrible. The next row is a visible improvement. Although, if one runs such tests a few times per day, every day at a large scale, then there is still a small chance of a collision about once a year, I guess.

If it was just me, then I would go for one of the two latter schemes and call it done, because I don't have an emotional attachment to the adjective part.  But if you prefer, we can try the second scheme and see how it goes?

porridge (approved):

PR #685, opened 2026-04-29
fix: retry transient API errors in wait/poll loops

porridge:

All four `wait.Poll*` call sites treated any non-`NotFound` error from `Get` as a fatal poll error (`return false, err`), causing `PollUntilContext*` to abort immediately. Transient Kubernetes API errors (Internal, ServerTimeout, network blips) would terminate the poll loop well before the configured timeout was reached.

This changes all four sites to return `(false, nil)` on transient errors so polling continues until the actual timeout expires. Also:
- `kubernetes.WaitForSA` is transformed into `Harness.waitForSA` so that it can take advantage of the logger in Harness (which is its only user)
- `WaitForDelete` turned out to be unused and was replaced with a very similar but subtly different w.r.t. used types implementation from `Step.DeleteExisting`
- Adds tests for the new `WaitForDelete`.

### DIFF
--- a/internal/harness/harness.go
+++ b/internal/harness/harness.go
@@ -17,8 +17,11 @@ import (
 
 	docker "github.com/moby/moby/client"
 	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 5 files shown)


vladbologa (changes requested):
I think it would be useful to surface somehow what the actual error was. Right now all we'd get is `DeadlineExceeded`, right?

vladbologa (commented):
I think it would be useful to surface somehow what the actual error was. Right now all we'd get is `DeadlineExceeded`, right?

vladbologa (approved):

porridge:
@vladbologa good point, I also noticed `kubernetes.WaitForDelete` was unused so did some shuffling around such that all cases surface the transient errors in one way or another, at least when they cause an eventual failure.
PTAL

vladbologa (approved):

PR #709, opened 2026-08-13
refactor: avoid reprocessing in LoadYAML

### LINKED ISSUE #643: Refactor Step.LoadYAML
On every file load, this function re-processes all objects loaded into the step so far.

This pure waste.

Green00101:

This PR avoids reprocessing objects in `step.LoadYAML`, and adds a test.

Fixes #643

### DIFF
--- a/internal/step/step.go
+++ b/internal/step/step.go
@@ -536,13 +536,16 @@ func (s *Step) LoadYAML(f kfile.Info) error {
 		return err
 	}
 
+	assertStart := len(s.Asserts)
+	applyStart := len(s.Apply)
+
 	if err = s.populateObjectsByType(f, objects); err != nil {
 		return fmt.Errorf("populating step: %v", err)
 	}
 
-	asserts := []client.Object{}
+	asserts := s.Asserts[:assertStart]
 
-	for _, obj := range s.Asserts {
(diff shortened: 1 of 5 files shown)


internal/step/test_data/load_yaml_once/shared/pod.yaml
porridge, line 1: Please add trailing newlines to this and other YAML files.

internal/step/step.go
```diff
@@ -536,13 +536,16 @@ func (s *Step) LoadYAML(f kfile.Info) error {
 		return err
 	}
 
+	assertStart := len(s.Asserts)
+	applyStart := len(s.Apply)
```
porridge, line 540: I think rather than keep appending new objects into fields of `s` and then try to skip the old ones, we should refactor this method to process newly loaded objects first, and only append them to fields of `s` afterwards.
There's **a lot** of complexity going on here (e.g. the `TestStep` object processing and how treatment of resources loaded by it deviate in subtle ways from those loaded directly), and adding more variables makes it even more convoluted.

porridge (changes requested):
Wow, I didn't even realize there were visible side effects from this repeated processing!

PR #715, opened 2026-08-31
add: new `--delete` flag

### LINKED ISSUE #708: Add a new flag for skipping deletion of namespaces for failed tests
Currently, we have the `--skip-delete` flag on KUTTL, which skips the deletion of **allI** created resources during the tests. I would like to have something similar, but specify to KUTTL to skip the deletion only to namespaces with failing tests.

**Why is this needed**:
The main purpose is debugging and troubleshooting. In [ORC](https://github.com/k-orc/openstack-resource-controller/), our CI collects the resources managed by our controller when the CI step fails for further debugging. The default behavior is to delete all resources after completion, and this makes our CI to collect nothing when facing a failure. See https://github.com/k-orc/openstack-resource-controller/issues/855.

Using the `--skip-delete` partially works for us. While it avoids the resources deletion, we accumulate resources during the run. Besides, we rely on a deletion mechanism to make sure that some resources are in fact deleted.

If there is any other way to reach this ☝️, let me know.

winiciusallan:
This PR introduces a new flag to kuttl in which the user is now able to select which resources will be deleted after the run. This is specially useful when you would like to debug your tests by keeping some resources alive.

Currently, kuttl has the `--skip-delete` flag which skips deletion for all resources regardless it's passed or not. 

This PR also marks the `--skip-delete` flag as deprecated since we can have the same behavior by setting `--delete none`.

Fixes #708

### DIFF
--- a/internal/harness/harness.go
+++ b/internal/harness/harness.go
@@ -87,7 +87,7 @@ func (h *Harness) LoadTests(dir string) ([]*testcase.Case, error) {
 		tests = append(tests, testcase.NewCase(
 			dirEntry.Name(),
 			dir,
-			testcase.WithSkipDelete(h.TestSuite.SkipDelete),
+			testcase.WithDeletePolicy(h.TestSuite.ResolvedDeletePolicy()),
 			testcase.WithNamespace(h.TestSuite.Namespace),
 			testcase.WithTimeout(timeout),
 			testcase.WithLogSuppressions(h.TestSuite.Suppress),
--- a/internal/kuttlctl/cmd/test.go
+++ b/internal/kuttlctl/cmd/test.go
@@ -55,6 +55,7 @@ func newTestCmd() *cobra.Command { //nolint:gocyclo
 	kindConfig := ""
 	kindContext := ""
 	skipDelete := false
+	deletePolicy := ""
 	skipClusterDelete := false
 	parallel := 0
 	artifactsDir := ""
@@ -166,11 +167,23 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
 			// perhaps there are cases where that is part of the test.  In general, there is no cluster to delete and
 			// there is no namespace controller.
 			if options.StartControlPlane {
-				options.SkipDelete = true
+				options.Delete = harnessApi.DeleteNone
 			}
 
 			if isSet(flags, "skip-delete") {
-				options.SkipDelete = skipDelete
+				log.Println("--skip-delete is deprecated; use --delete=none instead")
+				if skipDelete {
+					options.Delete = harnessApi.DeleteNone
+				}
+			}
+
+			if isSet(flags, "delete") {
+				switch harnessApi.DeletePolicy(deletePolicy) {
+				case harnessApi.DeleteAll, harnessApi.DeleteSuccess, harnessApi.DeleteNone:
+					options.Delete = harnessApi.DeletePolicy(deletePolicy)
+				default:
+					return fmt.Errorf("invalid --delete value %q: must be one of all, success, none", deletePolicy)
+				}
 			}
 
 			if isSet(flags, "skip-cluster-delete") {
@@ -271,7 +284,8 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
 	testCmd.Flags().StringVar(&kindConfig, "kind-config", "", "Specify the KIND configuration file path (implies --start-kind, cannot be used with --start-control-plane).")
 	testCmd.Flags().StringVar(&kindContext, "kind-context", "", "Specify the KIND context name to use (default: kind).")
 	testCmd.Flags().StringVar(&artifactsDir, "artifacts-dir", "", "Directory to output kind logs to (if not specified, the current working directory).")
-	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "If set, do not delete resources created during tests (helpful for debugging test failures, implies --skip-cluster-delete).")
+	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "Deprecated: use --delete=none instead. If set, do not delete resources created during tests (helpful for debugging test failures, implies --skip-cluster-delete).")
+	testCmd.Flags().StringVar(&deletePolicy, "delete", "", "Control which resources are deleted after tests: all (default), success (only on pass), none (never delete, useful for debugging).")
 	testCmd.Flags().BoolVar(&skipClusterDelete, "skip-cluster-delete", false, "If set, do not delete the mocked control plane or kind cluster.")
 	// The default value here is only used for the help message. The default is actually enforced in RunTests.
 	testCmd.Flags().IntVar(&parallel, "parallel", 8, "The maximum number of tests to run at once.")
--- a/internal/kuttlctl/cmd/test.go
+++ b/internal/kuttlctl/cmd/test.go
@@ -178,10 +178,12 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
 			}
 
 			if isSet(flags, "delete") {
-				switch harnessApi.DeletePolicy(deletePolicy) {
-				case harnessApi.DeleteAll, harnessApi.DeleteSuccess, harnessApi.DeleteNone:
+				if isSet(flags, "skip-delete") {
+					return fmt.Errorf("you can not set both --skip-delete and --delete; use only --delete instead")
+				}
+				if harnessApi.ValidDeletePolicy(harnessApi.DeletePolicy(deletePolicy)) {
 					options.Delete = harnessApi.DeletePolicy(deletePolicy)
-				default:
+				} else {
 					return fmt.Errorf("invalid --delete value %q: must be one of all, success, none", deletePolicy)
 				}
 			}
@@ -284,8 +286,8 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
 	testCmd.Flags().StringVar(&kindConfig, "kind-config", "", "Specify the KIND configuration file path (implies --start-kind, cannot be used with --start-control-plane).")
 	testCmd.Flags().StringVar(&kindContext, "kind-context", "", "Specify the KIND context name to use (default: kind).")
 	testCmd.Flags().StringVar(&artifactsDir, "artifacts-dir", "", "Directory to output kind logs to (if not specified, the current working directory).")
-	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "Deprecated: use --delete=none instead. If set, do not delete resources created during tests (helpful for debugging test failures, implies --skip-cluster-delete).")
-	testCmd.Flags().StringVar(&deletePolicy, "delete", "", "Control which resources are deleted after tests: all (default), success (only on pass), none (never delete, useful for debugging).")
+	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "Deprecated: same as --delete=none - use that instead.")
+	testCmd.Flags().StringVar(&deletePolicy, "delete", "", "Control which resources are deleted after tests: all (default), success (only in passing test cases), none (never delete, useful for debugging).")
 	testCmd.Flags().BoolVar(&skipClusterDelete, "skip-cluster-delete", false, "If set, do not delete the mocked control plane or kind cluster.")
 	// The default value here is only used for the help message. The default is actually enforced in RunTests.
 	testCmd.Flags().IntVar(&parallel, "parallel", 8, "The maximum number of tests to run at once.")
--- a/internal/step/delete_policy_test.go
+++ b/internal/step/delete_policy_test.go
@@ -1,0 +1,115 @@
+package step
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/kudobuilder/kuttl/internal/kubernetes"
+	k8sfake "github.com/kudobuilder/kuttl/internal/kubernetes/fake"
+	testutils "github.com/kudobuilder/kuttl/internal/utils"
+	harness "github.com/kudobuilder/kuttl/pkg/apis/testharness/v1beta1"
+)
+
+// newStepWithClient builds a minimal Step wired to the supplied fake client.
+func newStepWithClient(t *testing.T, cl client.Client, policy harness.DeletePolicy) *Step {
+	t.Helper()
+	return &Step{
+		DeletePolicy: policy,
+		Logger:       testutils.NewTestLogger(t, ""),
+		Client:       func(bool) (client.Client, error) { return cl, nil },
+		DiscoveryClient: func() (discovery.DiscoveryInterface, error) {
+			return k8sfake.DiscoveryClient(), nil
+		},
+		Apply: []client.Object{
+			kubernetes.NewPod("test-pod", ""),
+		},
+	}
+}
+
+// TestCreateDeletePolicy_None verifies that DeleteNone registers no cleanup callbacks,
+// leaving created objects alive after the test ends.
+func TestCreateDeletePolicy_None(t *testing.T) {
+	cl := fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+	s := newStepWithClient(t, cl, harness.DeleteNone)
+
+	errs := s.Create(t, testNamespace)
+	require.Empty(t, errs)
+
+	// Object was created and must survive – no cleanup is registered for DeleteNone.
+	pod := kubernetes.NewPod("test-pod", testNamespace)
+	require.NoError(t, cl.Get(t.Context(), kubernetes.ObjectKey(pod), pod))
+}
+
+// TestCreateDeletePolicy_All verifies that DeleteAll cleans up resources regardless of
+// whether the step succeeded.
+func TestCreateDeletePolicy_All(t *testing.T) {
+	cl := fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+	s := newStepWithClient(t, cl, harness.DeleteAll)
+	// s.succeeded remains false – DeleteAll must ignore it.
+
+	errs := s.Create(t, testNamespace)
+	require.Empty(t, errs)
+
+	pod := kubernetes.NewPod("test-pod", testNamespace)
+	require.NoError(t, cl.Get(t.Context(), kubernetes.ObjectKey(pod), pod), "object should exist right after Create")
+
+	// Simulate what the registered t.Cleanup would do via Clean().
+	require.NoError(t, s.Clean(testNamespace))
+	assert.True(t, k8serrors.IsNotFound(cl.Get(t.Context(), kubernetes.ObjectKey(pod), pod)),
+		"DeleteAll: object should be gone after cleanup")
+}
+
+// TestCreateDeletePolicy_Success_OnSuccess verifies that DeleteSuccess deletes resources
+// when s.succeeded is true at cleanup time.
+func TestCreateDeletePolicy_Success_OnSuccess(t *testing.T) {
+	var cl client.Client
+
+	t.Run("inner", func(t *testing.T) {
+		cl = fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+		s := newStepWithClient(t, cl, harness.DeleteSuccess)
+		s.succeeded = true // simulate a passing step
+
+		errs := s.Create(t, testNamespace)
+		require.Empty(t, errs)
+
+		pod := kubernetes.NewPod("test-pod", testNamespace)
+		require.NoError(t, cl.Get(t.Context(), kubernetes.ObjectKey(pod), pod))
+		// t.Cleanup fires when the sub-test ends; succeeded==true → deletion runs.
+	})
+
+	// After the sub-test cleanup the object must be gone.
+	pod := kubernetes.NewPod("test-pod", testNamespace)
+	assert.True(t, k8serrors.IsNotFound(cl.Get(t.Context(), kubernetes.ObjectKey(pod), pod)),
+		"DeleteSuccess+succeeded=true: object should be deleted after test cleanup")
+}
+
+// TestCreateDeletePolicy_Success_OnFailure verifies that DeleteSuccess preserves resources
+// when s.succeeded is false at cleanup time.
+func TestCreateDeletePolicy_Success_OnFailure(t *testing.T) {
+	var cl client.Client
+
+	t.Run("inner", func(t *testing.T) {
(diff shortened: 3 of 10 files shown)


internal/kuttlctl/cmd/test.go
```diff
@@ -166,11 +167,23 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
...
-				options.SkipDelete = skipDelete
+				log.Println("--skip-delete is deprecated; use --delete=none instead")
+				if skipDelete {
+					options.Delete = harnessApi.DeleteNone
+				}
+			}
+
+			if isSet(flags, "delete") {
+				switch harnessApi.DeletePolicy(deletePolicy) {
+				case harnessApi.DeleteAll, harnessApi.DeleteSuccess, harnessApi.DeleteNone:
```
porridge: This list is also present in `ResolvedDeletePolicy`. Please extract to a `ValidPolicy() bool` helper, or similar, so it's easier to extend the list of possible policies in the future.

internal/kuttlctl/cmd/test.go
```diff
@@ -166,11 +167,23 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
...
 
 			if isSet(flags, "skip-delete") {
-				options.SkipDelete = skipDelete
+				log.Println("--skip-delete is deprecated; use --delete=none instead")
+				if skipDelete {
+					options.Delete = harnessApi.DeleteNone
+				}
+			}
+
+			if isSet(flags, "delete") {
```
porridge, line 180: I think we should also reject both `--delete` and `--skip-delete` being set. Or at least if they are set to conflicting values.

internal/testcase/case.go
```diff
@@ -38,9 +38,21 @@ type getDiscoveryClientFuncType func() (discovery.DiscoveryInterface, error)
 type CaseOption func(*Case)
 
 // WithSkipDelete sets whether to skip deletion of resources.
+// Deprecated: use WithDeletePolicy instead.
```
porridge: Since this is in `internal/` I think we can just delete it, no?

internal/kuttlctl/cmd/test.go
```diff
@@ -271,7 +284,8 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
 	testCmd.Flags().StringVar(&kindConfig, "kind-config", "", "Specify the KIND configuration file path (implies --start-kind, cannot be used with --start-control-plane).")
 	testCmd.Flags().StringVar(&kindContext, "kind-context", "", "Specify the KIND context name to use (default: kind).")
 	testCmd.Flags().StringVar(&artifactsDir, "artifacts-dir", "", "Directory to output kind logs to (if not specified, the current working directory).")
-	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "If set, do not delete resources created during tests (helpful for debugging test failures, implies --skip-cluster-delete).")
+	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "Deprecated: use --delete=none instead. If set, do not delete resources created during tests (helpful for debugging test failures, implies --skip-cluster-delete).")
```
porridge:
```suggestion
	testCmd.Flags().BoolVar(&skipDelete, "skip-delete", false, "Deprecated: same as --delete=none - use that instead.")
```

internal/kuttlctl/cmd/test.go
```diff
@@ -271,7 +284,8 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
...
+	testCmd.Flags().StringVar(&deletePolicy, "delete", "", "Control which resources are deleted after tests: all (default), success (only on pass), none (never delete, useful for debugging).")
```
porridge:
```suggestion
	testCmd.Flags().StringVar(&deletePolicy, "delete", "", "Control which resources are deleted after tests: all (default), success (only in passing test cases), none (never delete, useful for debugging). Values other than all imply --skip-cluster-delete).")
```
The question is: do they really imply --skip-cluster-delete? I wasn't able to see the connection, did I miss it?

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -68,7 +80,10 @@ type TestSuite struct {
 	// Containers to load to each KIND node prior to running the tests.
 	KINDContainers []string `json:"kindContainers"`
 	// If set, do not delete the resources after running the tests (implies SkipClusterDelete).
+	// Deprecated: use Delete instead.
 	SkipDelete bool `json:"skipDelete"`
+	// Delete controls which resources are deleted after a test run.
```
porridge, line 86: Please also update these descriptions similar to suggested CLI flag description changes.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -12,6 +12,18 @@ const KubeconfigLoadingEager = "Eager"
...
 
+// DeletePolicy controls which test resources are deleted after a test run.
+type DeletePolicy string
+
+const (
+	// DeleteAll deletes all resources after the test, regardless of pass or fail (default).
+	DeleteAll DeletePolicy = "all"
+	// DeleteSuccess deletes resources only when the test case passes without errors.
+	DeleteSuccess DeletePolicy = "success"
+	// DeleteNone skips deletion of all resources.
```
porridge:
```suggestion
	// DeleteNone skips deletion of resources in all test cases.
```

internal/testcase/case.go
```diff
@@ -297,12 +317,16 @@ func (c *Case) maybeReportEvents() {
 func (c *Case) Run(test *testing.T, rep report.TestReporter) {
 	defer rep.Done()
 
+	// caseSucceeded is set to true once every step passes; it gates DeleteSuccess cleanups.
+	caseSucceeded := false
```
porridge: Since every user of this variable also seems to have access to `c`, perhaps we can just make it a field of `Case` and avoid the plumbing of the pointer? 🤔 Similar to what you did for `Step`.

internal/testcase/case.go
```diff
@@ -297,12 +317,16 @@ func (c *Case) maybeReportEvents() {
...
+
 	setupReport := rep.Step("setup")
-	if err := c.setup(test); err != nil {
+	if err := c.setup(test, &caseSucceeded); err != nil {
```
porridge: What if namespace creations failed in some contexts, but not others, and setup - and thus the whole case - failed? It think it would make sense to behave consistently and retain the created ones if the policy is set to `DeleteSuccess`... WDYT? Or should we special case the namespace resources?

internal/step/delete_policy_test.go
```diff
@@ -0,0 +1,115 @@
...
+// TestCreateDeletePolicy_None verifies that DeleteNone registers no cleanup callbacks,
+// leaving created objects alive after the test ends.
+func TestCreateDeletePolicy_None(t *testing.T) {
+	cl := fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+	s := newStepWithClient(t, cl, harness.DeleteNone)
+
+	errs := s.Create(t, testNamespace)
+	require.Empty(t, errs)
+
+	// Object was created and must survive – no cleanup is registered for DeleteNone.
```
porridge: Um, must survive... what? This test does not seem to run anything that touches object deletion, like the other cases do with `Clean()` or a sub-case...

porridge (changes requested):
Looks very nice!
A bunch of requests, questions and nitpicks inline.
Please also remember to update the docs in the `docs/` directory!

internal/step/delete_policy_test.go
winiciusallan: yes, it does not, but it is testing that, with DeleteNone policy, no clear callback was registered and the resource remains existing in the cluster.

internal/testcase/case.go
winiciusallan: Maybe this is a lack of knowing how kuttl behaves when running a case, but correct if I'm wrong.

It runs each case concurrently, and each case runs on their own and unique k8s namespace. If namespace creation fails

```go
if err := c.setup(test, &caseSucceeded); err != nil {
	setupReport.Failure(err.Error())
	test.Fatal(err)
}
```

The case's goroutine will exit in `test.Fatal(err)` and on this case there is nothing to retain. When another case will run, it will create its namespace and its resource, run the steps, and then cleanup based on the passed policy.

So, I think a namespace creation failure is independent between cases, and thus we don't need to worry about retaining or not?

internal/testcase/case.go
winiciusallan: Yup, I just did. PTAL

pkg/apis/testharness/v1beta1/test_types.go
winiciusallan: Sorry, didn't get this exactly. Do you mean by what was suggested with this comment?

https://github.com/kudobuilder/kuttl/pull/715#discussion_r3911418990

internal/kuttlctl/cmd/test.go
winiciusallan: No, currently none of the `--delete` values are implying --skip-cluster-delete, which might be wrong since `--delete=none` should be the same as --skip-delete (and this indeed implies the cluster deletion).

I think we should do this ^, but I have a question: for `--delete=success`, should we handle the cluster deletion at runtime depending on if a test has failed or not? In a better language:

success & caseSucceeded; then deleteCluster
success & !caseSucceeded; then skipDeleteCluster?

I haven't done the changes yet, I'll wait for a +1 from your side.

internal/testcase/case.go
winiciusallan: Yes, I believe we can. I missed that. The only remaining reference was in `internal/testcase/case_integration_test.go` which I changed to use DeletePolicy("none") instead.

internal/kuttlctl/cmd/test.go
winiciusallan: Agreed. I do not see a reason why someone would like to set both, so I'm rejecting this case. LMK if you think the error message is good.

internal/kuttlctl/cmd/test.go
winiciusallan: I've created a `ValidDeletePolicy(dp DeletePolicy) bool` helper and changed other parts accordingly.

winiciusallan:
@porridge thanks for the thoroughly review :slightly_smiling_face:! it's been nice know more about the code base.

internal/step/delete_policy_test.go
porridge: > it is testing that [...] no clear callback was registered

**How** is it testing that? I must be missing something. I mean, if there _was__ a callback registered by mistake, this test would not have found about this. AFAICT it's checking the pod existence _before_ the callback would have a chance to run.

pkg/apis/testharness/v1beta1/test_types.go
porridge: Yes, exactly - enumerate the options.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -20,7 +20,7 @@ const (
 	DeleteAll DeletePolicy = "all"
 	// DeleteSuccess deletes resources only when the test case passes without errors.
 	DeleteSuccess DeletePolicy = "success"
-	// DeleteNone skips deletion of all resources.
+	// DeleteNone skips deletion of resources in all test cases.
 	DeleteNone DeletePolicy = "none"
 )
```
porridge, line 25: Please add a comment here reminding to keep `ValidDeletePolicy` in sync.

internal/testcase/case_test.go
```diff
@@ -500,8 +500,10 @@ func TestCase_createNamespace(t *testing.T) {
 				logger:         testutils.NewTestLogger(t, ""),
 			}
 
-			caseSucceeded := true // always treat as succeeded for these unit tests
-			gotErr := c.createNamespace(tm, clk, &caseSucceeded)
+			// Treat as succeeded for these unit tests (namespace deletion should proceed).
+			succeeded := true
+			c.succeeded = &succeeded
```
porridge, line 505: Please put these closer right after definition of `c`. Also I think you can do `new(true)` these days?

internal/testcase/case.go
```diff
@@ -136,6 +124,7 @@ type Case struct {
 	ns                 *namespace
 	getClient          getClientFuncType
 	getDiscoveryClient getDiscoveryClientFuncType
+	succeeded          *bool
```
porridge, line 127: Does this need to be a pointer? 🤔

internal/kuttlctl/cmd/test.go
porridge: I think if only `--delete=success` was set, cluster deletion should depend on case success. If also `--skip-cluster-delete` was also set, then cluster should be kept unconditionally.

pkg/apis/testharness/v1beta1/test_types.go
winiciusallan: Done. PTAL

internal/step/delete_policy_test.go
winiciusallan: OK, I stepped back and took a better look at how it's handling the tests. I changed the pattern to call `s.Create(t, namespace)` which indeed register the cleanup callback function depending on the defined policy.

I saw that the test for DeletePolicyAll was also not exercising the cleanup path.

pkg/apis/testharness/v1beta1/test_types.go
winiciusallan: Done!

internal/testcase/case_test.go
winiciusallan: since I changed to be a primitive value, I'm not using `new(true)`, but moved closer to `c`.

internal/testcase/case.go
winiciusallan: nope, it does not need. I changed to be a primitive boolean. turns out this is way simpler :smile:

internal/kuttlctl/cmd/test.go
winiciusallan: I exported the `succeeded` variable to the harness file since this is where the cluster deletion skip logic exists. Considering there is a lot of parallel tests that can run, I'm using the client lock that exists. Do you think this approach makes sense?

winiciusallan:
@porridge sorry for the delay in giving another round of commits on this. thanks for being responsive and for the valuable feedbacks!

I'm splitting the follow-up changes in new commits to make the review easier, but if you want to squash things before merging, just say the word.

internal/step/delete_policy_test.go
```diff
@@ -33,38 +33,45 @@ func newStepWithClient(t *testing.T, cl client.Client, policy harness.DeletePoli
...
 // leaving created objects alive after the test ends.
 func TestCreateDeletePolicy_None(t *testing.T) {
-	cl := fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
-	s := newStepWithClient(t, cl, harness.DeleteNone)
+	var cl client.Client
 
-	errs := s.Create(t, testNamespace)
-	require.Empty(t, errs)
+	t.Run("inner", func(t *testing.T) {
+		cl = fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
```
porridge: I think it would be best to initialize the client outside this closure.
Similarly in the test below.

internal/testcase/case_test.go
```diff
@@ -489,6 +489,7 @@ func TestCase_createNamespace(t *testing.T) {
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
 			c := NewCase(name, "", tt.options...)
+			c.succeeded = true
```
porridge: Please leave the `			// Treat as succeeded for these unit tests (namespace deletion should proceed).` comment for this line, I found it helpful.

pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -15,6 +15,8 @@ const KubeconfigLoadingLazy = "Lazy"
 // DeletePolicy controls which test resources are deleted after a test run.
 type DeletePolicy string
 
+// If you are adding a new delete policy, remember to also add it to
+// ValidDeletePolicy function at this same file.
```
porridge: Perhaps move this comment between the last value and the closing `)`? If this list gets longer, people might overlook a comment at the top.

internal/harness/harness.go
```diff
@@ -616,6 +623,29 @@ func (h *Harness) Stop() {
...
+	}
+	policy := h.TestSuite.ResolvedDeletePolicy()
+	if policy == harness.DeleteNone {
+		return true
+	}
+	if policy == harness.DeleteSuccess {
+		h.clientLock.Lock()
+		failed := h.anyTestFailed
+		h.clientLock.Unlock()
+		return failed
```
porridge: I think
```suggestion
		return h.T.Failed()
```
would allow us to avoid having to introduce `h.anyTestFailed`, no?

internal/harness/harness.go
```diff
@@ -616,6 +623,29 @@ func (h *Harness) Stop() {
...
+	}
+	return h.TestSuite.SkipClusterDelete
```
porridge: I think
```suggestion
	return false
```
is equivalent (since the true case is taken care of at the top) and less demanding cognitively...

internal/harness/harness.go
```diff
@@ -616,6 +623,29 @@ func (h *Harness) Stop() {
 	}
 }
 
+// shouldSkipClusterDelete returns true when the cluster should be kept alive after the run.
+// If --skip-cluster-delete was explicitly set by the user it always wins.
+// Otherwise the decision is inferred from the delete policy:
+// - DeleteNone always keep the cluster.
+// - DeleteSuccess keeps the cluster when at least one test case failed.
```
porridge: Maybe we can avoid rephrasing the exact implementation for a private method and just say:
```suggestion
// Otherwise the decision is inferred from the delete policy and test success/failure.
```

internal/testcase/case.go
```diff
@@ -165,6 +166,11 @@ func (c *Case) GetName() string {
 	return c.name
 }
 
+// Succeeded returns true if every step of the test case passed.
+func (c *Case) Succeeded() bool {
```
porridge: I think we can remove this method since it was only called by the harness which should now be able to use the testing framework directly?

internal/testcase/case.go
```diff
@@ -334,10 +348,16 @@ func (c *Case) Run(test *testing.T, rep report.TestReporter) {
 			for _, err := range errs {
 				test.Error(err)
 			}
+			caseFailed = true
```
porridge: I think we could get rid of this variable if we replaced the succeeded field with failed and this line with:

```suggestion
			c.failed = true
```

porridge (changes requested):
Thank you for taking the time to make the cluster deletion logic match the documented behaviour!

internal/testcase/case.go
porridge: The one namespace per case is the easy case. Note however that kuttl supports multiple-cluster tests with per-step kubeconfig contexts. For each case, namespaces (with the same name) are created for each context, and this can succeed in some of them but fail in others.

internal/kuttlctl/cmd/test.go
porridge: I think we can reuse `testing.T.Failed()`, see inline comment.

porridge (changes requested):
Please also remember to:
- fix the DCO (feel free to squash the commits now if this makes it easier to you), and
- update the docs in the docs/ directory!

internal/testcase/case.go
winiciusallan: Done!

internal/testcase/case.go
winiciusallan: Yup, we can. harness file is not using it anymore.

internal/harness/harness.go
winiciusallan: Done

internal/harness/harness.go
winiciusallan: Done

internal/harness/harness.go
winiciusallan: oh, I did not know about this field in the testing framework. ok, this helps a lot since the implementation is already concurrent-safe.

Did the changes accordingly.

pkg/apis/testharness/v1beta1/test_types.go
winiciusallan: Done

internal/testcase/case_test.go
winiciusallan: This is not applicable anymore since we dropped the `succeeded` field to use `c.failed` instead, but left a comment with the same logical effect.

internal/step/delete_policy_test.go
winiciusallan: Changed all the tests which uses this same closure to initialize the client at the start of the function.

winiciusallan:
> Please also remember to:
> 
>     * fix the DCO (feel free to squash the commits now if this makes it easier to you), and
> 
>     * update the docs in the docs/ directory!

I fixed all commits without a sign while adding a new one (https://github.com/kudobuilder/kuttl/pull/715/commits/9d232c3815b48ad7adbcf9cc2697fd18284f3ce8) for addressing your last comments. If you think this is now ready to go, I'll squash things into two commits: one for the implementation itself and another one for the documentation.

porridge:
@winiciusallan I'm OOO till Wednesday but I should be able to TAL on Thursday.

winiciusallan:
> @winiciusallan I'm OOO till Wednesday but I should be able to TAL on Thursday.

Thanks for letting me know! There is no rush :slightly_smiling_face:

internal/step/delete_policy_test.go
porridge, line 1: Please have a look at https://github.com/winiciusallan/kuttl/pull/1 which converts this into a table test.

internal/testcase/case.go
```diff
@@ -234,8 +238,12 @@ func (c *Case) createNamespace(test T, cl clientWithKubeConfig) error {
 	if err != nil {
 		return fmt.Errorf("failed to create test namespace %q: %w", c.ns.name, err)
 	}
-	if !c.skipDelete {
+	if c.deletePolicy != v1beta1.DeleteNone {
 		test.Cleanup(func() {
+			if c.deletePolicy == v1beta1.DeleteSuccess && c.failed {
+				cl.Logf("Skipping namespace deletion for %q: test did not pass (delete policy: success)", c.ns.name)
```
porridge:
```suggestion
				cl.Logf("Skipping deletion of namespace %q: test failed & delete policy=success", c.ns.name)
```

internal/testcase/case.go
```diff
@@ -334,6 +342,7 @@ func (c *Case) Run(test *testing.T, rep report.TestReporter) {
 			for _, err := range errs {
 				test.Error(err)
 			}
+			c.failed = true
```
porridge: Please put this right after the `if`. This will help ensure in the future that this flag is set as expected even if someone changes remaining code in this block in a way that might panic.

internal/testcase/case.go
porridge, line 311:
```suggestion
		c.failed = true
		setupReport.Failure(err.Error())
```

internal/testcase/case.go
porridge: I added a suggestion now that should address this.

porridge (changes requested):
We're getting there!

winiciusallan:
@porridge thanks for pushing the test table, it's indeed a best go practice. took the opportunity to squash the commits.

please see if you think it's now good.

porridge (approved):
❤️

PR #721, opened 2026-09-21
fix: respect case-insensitive reportFormat in TestSuite config

### LINKED ISSUE #449: reportFormat in TestSuite

porridge:
## What

Fixes #449: `reportFormat: XML` (or any non-lowercase value) in `kuttl-test.yaml` was ignored and a JSON report was produced instead, contrary to the documentation.

## Why

The `reportFormat` value from the config file was passed to the report writer verbatim, but the writer matched only lowercase `xml`/`json`, so uppercase/mixed-case values silently fell through to JSON. The `--report` CLI flag already lowercased its input, so only the config-file path was affected.

Note: normalization cannot be done at deserialization time. kuttl loads its config via k8s' unstructured converter, which does **not** honor a custom `UnmarshalJSON` on a string-based type (verified empirically). So the value is normalized explicitly in code, at the command-line boundary.

## What changed

- New typed `ReportType` in the API with `XML`/`JSON` constants and `Valid()`/`Normalize()` helpers; `TestSuite.ReportFormat` is now `ReportType`.
- The report format is normalized once at the command-line boundary (covers both the config file and `--report`).
- Unrecognized values now emit a **warning** and preserve the historical fallback based on where the effective value came from — the `--report` flag takes precedence over the config file:
  - from the flag → no report generated
  - from the config file → JSON report
- `report.Report` now owns a single format decision (`""` → no report, `XML`/`JSON` → write, anything else → error) with no silent JSON fallback.

## Behavior

"CLI flag" = `--report VALUE`; "Config field" = `reportFormat: VALUE` (flag unset).

### Before

| Value | CLI flag | Config field |
|---|---|---|
| `""` | none | none |
| `"   "` | none | JSON |
| `xml` | XML | XML |
| `XML` | XML | **JSON** ← bug |
| `json` | JSON | JSON |
| `JSON` | JSON | JSON |
| `foobar` | none *(silent)* | JSON *(silent)* |

### After

| Value | CLI flag | Config field |
|---|---|---|
| `""` | none | none |
| `"   "` | none (+ warning ⚠️) | JSON (+ warning ⚠️) |
| `xml` | XML | XML |
| `XML` | XML | **XML** ← fixed |
| `json` | JSON | JSON |
| `JSON` | JSON | JSON |
| `foobar` | none (+ warning ⚠️) | JSON (+ warning ⚠️) |

The only outcome change is `XML` in the config file (JSON → XML, the fix). `"   "` and `foobar` keep the same outcomes as before, now with a warning instead of silent behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/apis/testharness/v1beta1/test_types.go
```diff
@@ -102,9 +136,9 @@ type TestSuite struct {
 	// Commands to run prior to running the tests.
 	Commands []Command `json:"commands"`
 
-	// ReportFormat determines test report format (JSON|XML|nil) nil == no report
-	// maps to report.Type, however we don't want generated.deepcopy to have reference to it.
-	ReportFormat string `json:"reportFormat"`
+	// ReportFormat determines test report format (JSON|XML|"") "" == no report.
```
mclasmeier:
```suggestion
	// ReportFormat determines test report format (JSON|XML|""), where "" means no report.
```

pkg/apis/testharness/v1beta1/report_type_test.go
```diff
@@ -0,0 +1,49 @@
...
+		in       ReportType
+		expected ReportType
+	}{
+		"lowercase xml":          {in: "xml", expected: ReportTypeXML},
+		"uppercase XML":          {in: "XML", expected: ReportTypeXML},
+		"mixed case Xml":         {in: "Xml", expected: ReportTypeXML},
+		"lowercase json":         {in: "json", expected: ReportTypeJSON},
+		"uppercase JSON":         {in: "JSON", expected: ReportTypeJSON},
+		"mixed case Json":        {in: "jSoN", expected: ReportTypeJSON},
+		"spaces are not trimmed": {in: "  xml  ", expected: "  XML  "},
```
mclasmeier, line 20: I'd actually say space trimming would be a reasonable thing to expect from a such a normalization function. WDYT?

pkg/apis/testharness/v1beta1/report_type_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"empty stays empty":      {in: "", expected: ReportTypeNil},
+		"unknown value upper":    {in: "yaml", expected: "YAML"},
```
mclasmeier: nit, could also use "foo" here, because "yaml" might actually be something that could potentially be implemented one day and that would kinda invalidate this "unknown value" name.

same in the test below.

internal/report/report.go
```diff
@@ -341,22 +333,26 @@ func (ts *Testsuites) Close() {
 // latestEnd provides the time of the latest end out of the collection of testcases
 
 // Report prints a report for TestSuites to the directory.  ftype == json | xml.
```
mclasmeier: remove?

internal/report/report.go
```diff
@@ -341,22 +333,26 @@ func (ts *Testsuites) Close() {
...
-	err := ensureDir(dir)
-	if err != nil {
+	if err := ensureDir(dir); err != nil {
 		return err
 	}
 
-	// if a report is requested it is always created
-	switch ftype {
-	case XML:
+	switch ftype { //nolint:exhaustive // ReportTypeNil is handled by the early return above; any other value is an error.
```
mclasmeier: alternative would be to keep exhaustiveness and add a dummy entry for the `Nil` branch. This would allow static analyzers to support future extensibility. Just an idea, not requiring change.

internal/kuttlctl/cmd/test.go
```diff
@@ -197,8 +196,22 @@ For more detailed documentation, visit: https://github.com/kudobuilder/kuttl`,
...
+			// are matched case-insensitively for backward compatibility. See issue #449.
+			rawReportFormat := options.ReportFormat
+			options.ReportFormat = options.ReportFormat.Normalize()
+			if !options.ReportFormat.Valid() {
+				// Keep the (inconsistent) fallback behaviour for backward compatibility, but warn.
+				if isSet(flags, "report") {
+					log.Printf("Warning: unrecognized --report format %q; no report will be generated (expected one of JSON, XML)", rawReportFormat)
+					options.ReportFormat = harnessApi.ReportTypeNil
+				} else {
+					log.Printf("Warning: unrecognized report format %q; defaulting to JSON (expected one of JSON, XML)", rawReportFormat)
```
mclasmeier, line 212: I suppose there is no real "warn"-severity logging available, just prefixed printf?

mclasmeier (commented):
Mostly minor nits/questions.

I would suggest to add whitespace trimming in the normalization, so that accidentally introduced whitespaces are simply ignored.

One question I would like to ask though, is `ReportFormatNil` really the best name here? I think I would prefer something like `ReportFormatNone`.

mclasmeier (approved):
Feel free to merge once the comments are looked at.

internal/report/report.go
porridge: Then the `unreachable` linter is unhappy.

pkg/apis/testharness/v1beta1/report_type_test.go
porridge: I actually removed trimming on purpose. It was complicating the code somewhere, and I'm not very fond of such "convenience" transformations 😟

internal/kuttlctl/cmd/test.go
porridge: Logging in kuttl is not great 😟 
All the more that it's all built around the go test framework which gives us some things for free, but is quite limiting in other places.

